PR #6, opened 2022-01-07
Update start commands

### LINKED ISSUE #5: babel-node isn't good for production
We shouldn't be using babel-node as part of `npm start` since it isn't meant for production.

Instead we should build and run directly from the transpiled code.

Let's update `start` to do that, but then also let's create a new `dev` command which runs through babel-node (and nodemon for good measure to restart on code change)

slifty:
This PR adds a `npm run dev` and also updates`npm start` to run off of transpiled code instead of through `babel-node`

To test this you need to run `npm install`

It also updates the build destination to be `/build` instead of `/lib` since that feels more explicit.

Resolves #5

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,4 +104,4 @@ dist
 .tern-port
 
 # Ignore transpiled code
-/lib
+/build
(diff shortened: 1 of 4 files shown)


cecilia-donnelly (approved):
Thanks @slifty !

PR #7, opened 2022-01-07
Add Sentry

### LINKED ISSUE #4: Sentry Integration
We still use Sentry, right?

Let's integrate with that similar to how we do over in [upload-service](https://github.com/PermanentOrg/upload-service)

slifty:
This PR makes it possible to configure an environment to report to Sentry

Testing this involves `npm install`

Resolves #4 

(This builds on top of #6, so please don't review until that is merged and this is rebased)


cecilia-donnelly (approved):
Looks good - I was able to fill out the .env and start the service, though we're not sending anything to Sentry yet so I couldn't confirm that it arrived.

Issue #3, opened 2022-01-07
Set up SSH Service


slifty:
SFTP is a protocol that runs over SSH2.

We don't have to implement an SSH2 server from scratch, since the [ssh2 npm module](https://www.npmjs.com/package/ssh2) exists.

This issue includes setting up the main application tooling and handling (at least) the basic SSH2 connectivity which includes properly handling the following events:

- [x] authentication (username + password; can be a static username / pass for now, as permanent auth integration will come next)
- [x] ready
- [x] close

More may be discovered through implementation of the above.  There is an [example SSH2 server provided by the module documents](https://www.npmjs.com/package/ssh2#password-and-public-key-authentication-and-non-interactive-exec-command-execution) though this should ONLY be used for reference as it does both more and less than what we want.

Success is the ability to invoke ssh, pass in a valid username and password, and have an open ssh2 connection with the server.

slifty:
I'm going to mark this as closed!

The service is indeed "set up" -- a lot needs to be added, but the scope of this issue is complete.

PR #8, opened 2022-01-17
Add stubbed SSH event handlers

slifty:
This PR sets up the skeleton of the SSH2 server aspect of our SFTP server.

These event handlers don't actually perform any logic, but they create spaces to cover all of the connection events defined in the [ssh2 package documentation](https://www.npmjs.com/package/ssh2#connection-events)

You can test if this is working by updating your `.env` with the new values, and then running:

```
ssh 127.0.0.1 -p22222
```

You should then see:

```
Unable to negotiate with 127.0.0.1 port 22222: no matching host key type found. Their offer:
```

^ Note that I am using `SSH_PORT=222222` and if you use a different value you'd need to adjust the test accordingly.

Issue #3

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -16,3 +16,11 @@ NODE_ENV=${NODE_ENV}
 # See: https://github.com/winstonjs/winston#logging-levels
 # e.g. error, warn, info, http, verbose, debug, silly
 LOG_LEVEL=${LOG_LEVEL}
+
+# The port the SSH service will listen on
+# See https://nodejs.org/api/net.html#serverlistenoptions-callback
+SSH_PORT=${SSH_PORT}
+
+# The hostname the SSH service will listen on
+# See https://nodejs.org/api/net.html#serverlistenoptions-callback
+SSH_HOST=${SSH_HOSTNAME}
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,4 +1,7 @@
 import * as Sentry from '@sentry/node';
+import { server } from './server';
+import { logger } from './logger';
+import type { ListenOptions } from 'net';
 
 if ('SENTRY_DSN' in process.env
  && 'SENTRY_ENVIRONMENT' in process.env) {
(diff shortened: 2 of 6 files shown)


src/classes/SshConnectionSession.ts
```diff
@@ -0,0 +1,23 @@
+export class SshConnectionSession {
```
slifty, line 1: This name is actually a bit overloaded, since SSH2 has the concept of a "connection" and a connection can have more than one "session" (see the `onSession` event handler defined in this class).

This class makes it seem like it will deal with the *session* but in fact it really is just intended to handle a connection for a given client.

Will think about names, but maybe something like `SshClientConnectionHandler`

src/index.ts
```diff
@@ -8,3 +10,15 @@ if ('SENTRY_DSN' in process.env
...
+
+const listenOptions: ListenOptions = {
+  port: Number.parseInt(process.env.SSH_PORT ?? '22'),
+  host: process.env.SSH_HOST ?? '127.0.0.1',
+};
+
+server.listen(
+  listenOptions,
+  function () {
+    console.log(`Listening for SSH requests on ${listenOptions.host ?? ''}:${listenOptions.port ?? ''}`);
```
slifty, line 22: We should probably use a real logger.

I'll set that up in another MR / issue.

slifty: Opened in #11 -- if that gets merged before this I'll switch to it (that'd probably be the ideal order)

Fenn-CS (changes requested):
As you said, it would be a good thing to start using that logger here already, the stubs make general sense to me. "Changes" to include the new logger in #11 after merge.

.env.example
```diff
@@ -7,3 +7,11 @@ SENTRY_DSN=${SENTRY_DSN}
 # See https://docs.sentry.io/product/sentry-basics/environments/
 # e.g. localhost
 SENTRY_ENVIRONMENT=${SENTRY_ENVIRONMENT}
+
+# The port the SSH service will listen on
+# See https://nodejs.org/api/net.html#serverlistenoptions-callback
+SSH_PORT=${SSH_PORT}}
```
cecilia-donnelly, line 13: Just a lil typo - extra bracket here.

cecilia-donnelly (approved):
Fun!  Thanks for the clear testing instructions, @slifty .

I am on node 14 locally and got an error installing `ssh2` around crypto building, though eventually npm told me "Succeeded in building optional crypto binding" so presumably there's a workaround.  Didn't cause any trouble in the end, but thought I should note it here.

.env.example
slifty: Great catch thank you

slifty:
I think we're about to merge logging so once that's done I'll rebase, use it, and merge this in.

Interesting regarding the node 14 crypto message... If we see that on 16 I'll open an issue and see what needs to be done about it.

slifty:
One note: this doesn't include tests though we probably will want them.  I know we should normally build tests into PRs so we don't accumulate debt.  Since this really is just stubbing rather than adding functionality I don't think tests make sense yet.

PR #9, opened 2022-01-17
Support host keys

slifty:
This PR adds documentation for creating / configuring the server with a host key, and adds logic for using that host key when starting the SSH2 server.

To test this you should:

1. Follow the new README instructions for generating a host key.
2. Populate the `.env` with a path to your new host key
3. Start the server.

Once you do that you should be able to run `ssh 127.0.0.1 -p22222` and see no warnings / etc.  It will sit there doing nothing:

image

When you stop the server you should see a "Connection closed" message on your SSH client (though this is probably client specific)


cecilia-donnelly (approved):
So cool!

PR #11, opened 2022-01-17
Configure logging

### LINKED ISSUE #10: Set up a real logger
Instead of using `console.log` let's use winston like we do in the [upload service](https://github.com/PermanentOrg/upload-service/blob/main/src/log.ts)

slifty:
This PR  practically copies the logger written for our [upload service](https://github.com/PermanentOrg/upload-service/issues/4).

The only change is that I directly installed the [logform](https://www.npmjs.com/package/logform) package.

Closes #10

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,8 @@
   },
   "dependencies": {
     "@sentry/node": "^6.16.1",
-    "dotenv": "^10.0.0"
+    "dotenv": "^10.0.0",
+    "logform": "^2.3.2",
+    "winston": "^3.4.0"
   }
 }
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,0 +1,49 @@
+import { createLogger, format, transports } from 'winston';
+import type { TransformableInfo } from 'logform';
+
+const {
+  colorize,
+  combine,
+  errors,
+  printf,
+  simple,
+} = format;
+
+const template = ({
+  level,
+  message,
+  stack,
+  ...rest
+}: TransformableInfo): string => {
+  const meta = Object.keys(rest).length > 0 ? ` ${JSON.stringify(rest)}` : '';
+  const stacktrace = typeof stack === 'string' ? `\n${stack}` : '';
+  const timestamp = new Date().toLocaleString();
+  return `${timestamp} ${level}: ${message}${meta}${stacktrace}`;
+};
+
+const debugLogger = createLogger({
+  format: combine(
+    errors({ stack: true }),
+    colorize(),
+    printf(template),
+  ),
+  level: process.env.LOG_LEVEL ?? 'debug',
+  transports: [
+    new transports.Console(),
+  ],
+});
+
+const prodLogger = createLogger({
+  format: combine(
+    errors({ stack: true }),
+    simple(),
+  ),
+  level: process.env.LOG_LEVEL ?? 'info',
+  transports: [
+    new transports.Console(),
+  ],
+});
+
+const logger = process.env.NODE_ENV === 'production' ? prodLogger : debugLogger;
+
+export { logger };
--- a/.env.example
+++ b/.env.example
@@ -7,3 +7,12 @@ SENTRY_DSN=${SENTRY_DSN}
 # See https://docs.sentry.io/product/sentry-basics/environments/
 # e.g. localhost
 SENTRY_ENVIRONMENT=${SENTRY_ENVIRONMENT}
+
+# The node environment this service is being run in.
+# e.g. production, testing, or development
+NODE_ENV=${NODE_ENV}
+
+# The npm log level that you want the logger to emit.
+# See: https://github.com/winstonjs/winston#logging-levels
+# e.g. error, warn, info, http, verbose, debug, silly
+LOG_LEVEL=${LOG_LEVEL}
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,9 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@sentry/node": "^6.16.1",
-        "dotenv": "^10.0.0"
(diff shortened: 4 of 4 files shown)


src/logger.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const debugLogger = createLogger({
+  format: combine(
+    errors({ stack: true }),
+    colorize(),
+    printf(template),
+  ),
+  level: process.env.LOG_LEVEL ?? 'debug',
+  transports: [
+    new transports.Console(),
```
Fenn-CS, line 32: Would be nice if `winston.transports.File` is used instead as opposed to console. In addition a `.logs` folder may be configured and tracked but it's content ignored in `.gitignore`. 

Otherwise, LGTM.

Fenn-CS (approved):

src/logger.ts
slifty: Great idea!

Do you think there would be utility on having logging got to console when in a development environment and to a log file in prod, or is that an antipattern?

Fenn-CS: That make sense! Besides you can make that a `.env` setting with a default. However inconvenient/convenient the developer would be able to change it!

Fenn-CS: So it would be added in `.env.example`.

slifty: Also cc @jasonaowen since he originally set up the logging config that this came from, and may have thoughts to add.

cecilia-donnelly (changes requested):
Thanks @slifty and @Fenn-CS!  Love the logger options.

Can you update .env.example with a NODE_ENV?

src/logger.ts
jasonaowen: My understanding, based on my research at the time, is that saving logs to files is a separate concern that should be handled out-of-application. upload-service and notification-service are started by systemd, and systemd collects their logs and exposes them via journalctl.

If the application manages its own logs, I think we would also need to implement log rotation, configuration for log rotation, configuration for the target files, and probably other things that I'm not thinking of now.

Can you say more about why you want to log to a file instead of the console, @Fenn-CS? What problem are you trying to solve?

package.json
```diff
@@ -51,6 +51,8 @@
   },
   "dependencies": {
     "@sentry/node": "^6.16.1",
-    "dotenv": "^10.0.0"
+    "dotenv": "^10.0.0",
+    "logform": "^2.3.2",
```
jasonaowen, line 55: Wait, why didn't I explicitly depend on logform in upload-service? This is a good change - nice catch, @slifty!

jasonaowen (approved):
This looks good to me, @slifty!

I agree with @cecilia-donnelly's request to update `.env.example`, and suspect that now that your attention has been drawn to it I don't need to explicitly write this, but please also add `LOG_LEVEL` to it. Approving because you don't otherwise need to block on me - once @cecilia-donnelly is satisfied, so am I.

package.json
slifty: @jasonaowen I was wondering that too but not enough to raise an issue at that point 😂

src/logger.ts
Fenn-CS: The logs are useful if someone can see them later, for production what is the "console"? is someone sitting there and watching?

putting the events in files means that they are captured when they happen and someone can see them later. If there's some other mechanism that ensure the same result is achieve then that's great.

However, it very common to see logging being handled as part of the application and not out that sounds like over engineering because they other log levels, like server logs before application logs.

slifty: So to answer the prod question, I think that console is generally piped to a file by the production deployment mechanism (and that seems to be a pretty standard / expected practice).  Some examples of production logging-to-console being the expected:

* Heroku logs are explicitly [the console output of all processes](https://devcenter.heroku.com/articles/logging)
* Supervisord also defines logging as one of it's primary functions, and in particular it [logs the output of child processes to whatever destination you specify](http://supervisord.org/logging.html).
* Docker seems to [emit logs to stdout and stderr](https://www.datadoghq.com/blog/docker-logging/).

Note: I don't remember what we're using to deploy these things right now, though I know it isn't Docker or Heroku!  But is it Supervisor?  Systemd?  Systemd is probably like supervisor in its log expectations.

---

Curve ball while we think about production logging: what if instead of file logging for production we consider defining a *Sentry* transport stream so logs are written to Sentry?  I think we're trying to use Sentry for Permanent?

---

All of that said: do you think we would want file logging in a dev environment?  I think that's more likely to be used than file-logging-from-application in our production environment, though it does represent complexity / a piece of the system we will have to maintain and update over time, so if console is good enough I think it's reasonable to stick with it.

Additional note: There is a [daily rotate transport](https://github.com/winstonjs/winston-daily-rotate-file) which will do log rotation, but I do agree that there's meaningful simplicity in NOT having our application have to also know how to manage log files.  Then again it's not great to have an application with a built-in time bomb that the engineer needs to actively diffuse (by setting up log rotation manually).

slifty: I should add this critical point: I think this decision should apply to *all our services* -- we should log in a consistent way; would it be appropriate to move forward on this PR (which mirrors our other services log styles), and to explore how we want service logging to work more broadly?

cecilia-donnelly: Yes, let's move forward, @slifty .

Fenn-CS: Thanks for the detailed explaination, it all make sense when the a mechanism to manage logging without much effort.

slifty: Ty all!  And Thank you @Fenn-CS for raising the discussion.

slifty:
Both environment variables are documented and ready for re-review @cecilia-donnelly

cecilia-donnelly (approved):
Thanks @slifty !

PR #12, opened 2022-01-17
Support "authentication"

slifty:
This PR adds a handler for the `authentication` event, though the handler isn't very smart yet.  It tells the client to use a password and then accepts whatever password the client provides.

While I'd like to say this is a philosophical decision and social commentary on the dehumanizing impact of nations and borders, the reality is I want to confirm that the permanent SDK is ready for action before incorporating it into the authentication flow.

You'll know this is working if you get a password prompt after running

```
ssh 127.0.0.1 -p22222
```

You should also see the following output on the service console:

image

### DIFF
--- a/src/classes/SshConnectionHandler.ts
+++ b/src/classes/SshConnectionHandler.ts
@@ -7,7 +7,7 @@ export class SshConnectionHandler {
    * https://datatracker.ietf.org/doc/html/rfc4252#section-5
    */
   public onAuthentication = (authContext: AuthContext): void => {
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
I didn't get the debugging output but that might be something with my local setup.  I still get some errors with `npm install`, so going to try to fix that.  Otherwise looks totally reasonable!

cecilia-donnelly:
Okay I updated some packages and installed `cmake` and I get the logging output now!  Hurray!

PR #13, opened 2022-01-18
Stub out SFTP session handler

slifty:
This PR adds the stubbed out skeleton for the set of sftp events that we'll need to implement in order to support the sftp protocol.  This won't actually do anything yet and so I don't believe an SFTP client would have a good time trying to invoke anything, however you should see activity if you try:

image

### DIFF
--- a/src/classes/SftpStreamHandler.ts
+++ b/src/classes/SftpStreamHandler.ts
@@ -8,7 +8,7 @@ export class SftpStreamHandler {
    * https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.3
    */
   public openHandler = (): void => {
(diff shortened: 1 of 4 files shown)


meisekimiu (approved):
Got it working and was able to get the canonicalize path request to trigger! (I assume the other ones can't really be triggered until the service actually responds to some SFTP events).

We discussed on slack that maybe the logger calls should be set to the `verbose` setting, which I think makes more sense. I initially couldn't see the logged statements because I assumed that they'd show up under the "verbose" setting.

That being said, I looked into this and these levels don't really have any formal definition other than "how important is this message on a scale from 0-5?" so as long as the levels we use in the application are consistent and we know we need to be in `debug` level logging to see future messages, I don't really see an issue.

PR #20, opened 2022-03-10
Document no support for symlinks

Fenn-CS:
Note our non-support for symlinks


Fenn-CS:
cc: @kfogel @slifty

slifty (approved):
Thank you for documenting this! Looks good.

PR #23, opened 2022-04-10
add definition for file & folder extensions

Fenn-CS:
This PR add the following to the collisions documentation;

- Definition of file extension
- New test cases covered
- Makes reference to the correct back-end issue, following up its implementation

### DIFF
--- a/docs/collisions.md
+++ b/docs/collisions.md
@@ -64,6 +64,54 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five 
 * [Extensions in folder names matter](https://apple.stackexchange.com/questions/123001/renamed-folder-becomes-a-file-with-an-extension) and should be handled :
 * Case sensitivity is taken into account. `Myfile` and `myFile` are different for example. That is useful to [ensure complete syncs](https://rclone.org/overview/#case-insensitive).
 
+#### Extensions
+
+For both files and folders the extension is any string that follows after the last dot.
+
+For example:
+<table>
+   <thead>
+      <th>File/Folder Name</th>
+      <th>Extension</th>
+      <th>Remark</th>
+   </thead>
+   <tbody>
+      <tr>
+         <td>My File.txt</td>
+         <td>txt</td>
+         <td>Regular</td>
+      </tr>
+      <tr>
+         <td>My File.txt(1)</td>
+         <td>txt(1)</td>
+         <td>Extension contains non-alphanumeric chars</td>
+      </tr>
+      <tr>
+         <td>My File.app</td>
+         <td>app</td>
+         <td>Regular</td>
+      </tr>
+      <tr>
+         <td>something.something</td>
+         <td>something</td>
+         <td>Name same as extension</td>
+      </tr>
+      <tr>
+         <td>2022.03.02</td>
+         <td>02</td>
+         <td>Extension is number</td>
+      </tr>
+      <tr>
+         <td>Folder.Work Files</td>
+         <td>Work Files</td>
+         <td>Extension contains spaces</td>
+      </tr>
+   </tbody>
+</table>
+
+As seen, extension can take infinite number of "shapes" BUT as long as all characters in the file name including the extension are valid file name chacters, then, **the extension is the string tha follows after the last dot** (Emphasis).
+
+
 ### Reserved/unsupported characters
 
 Characters that do not map to various file systems would be encoded. For example, `/` is not allowed in file names in all operating systems, while Windows goes ahead to restrict a lot more characters including `*`, `<`, `>`, `/`, `:`, `"` and `|`. 
@@ -394,7 +442,7 @@ NB: downloadFileName & downloadFolderName **should be recalculated after file or
    </td>
    <td><strong>In progress</strong>
    </td>
-   <td><a href="https://github.com/PermanentOrg/back-end/pull/11">https://github.com/PermanentOrg/back-end/pull/11</a>
+   <td><a href="https://github.com/PermanentOrg/back-end/pull/40">https://github.com/PermanentOrg/back-end/pull/40</a>
    </td>
   </tr>
   <tr>
@@ -402,7 +450,7 @@ NB: downloadFileName & downloadFolderName **should be recalculated after file or
    </td>
    <td><strong>In progress</strong>
    </td>
-   <td><a href="https://github.com/PermanentOrg/back-end/pull/11">https://github.com/PermanentOrg/back-end/pull/11</a>
+   <td><a href="https://github.com/PermanentOrg/back-end/pull/40">https://github.com/PermanentOrg/back-end/pull/40</a>
    </td>
   </tr>
   <tr>
@@ -418,10 +466,13 @@ NB: downloadFileName & downloadFolderName **should be recalculated after file or
 
 ## Testing Plan
 
-- Test generates unique & correctly formatted `downloadName` for n colliding files in the same namespace (parent folder).
-- Test generates unique & correctly formatted `downloadName` for n colliding folders in the same namespace.
-- Test generates unique & correctly formatted `downloadName` for n colliding files & folders in the same namespace.
-- Test generates unique & correctly formatted `downloadName` for incoming files and folders holding colliding deduplication strings.
+- Test generates directory-unique & correctly formatted `downloadName` for n colliding files in the same namespace (parent folder).
+- Test generates directory-unique & correctly formatted `downloadName` for n colliding folders in the same namespace.
+- Test generates directory-unique & correctly formatted `downloadName` for n colliding files & folders in the same namespace.
+- Test generates directory-unique & correctly formatted `downloadName` for incoming files and folders holding colliding deduplication strings.
+- Test generates directory-unique & correctly formatted `downloadName` for files and folders with uncommon naming styles.
+
+*Directory-unique: Download names need be unique only for files and folders in thesame directory*
 
 
 # Synchronization


docs/collisions.md
```diff
@@ -64,6 +64,61 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five
...
+      </tr>
+      <tr>
+         <td>Folder.Work Files</td>
+         <td>Work Files</td>
+         <td>Extension contains spaces</td>
+      </tr>
+   </tbody>
+</table>
+
+As seen, extension can take infinite number of "shapes" BUT as long as all characters in the file name including the extension are valid file name chacters, then, **the extension is the string tha follows after the last dot** (Emphasis).
```
kfogel, line 112: Typo: "tha" should be "that".

docs/collisions.md
```diff
@@ -64,6 +64,61 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five
...
+
+
+#### Source of extensions
+
+a) The **source string for file extensions ON FILES** is the **`uploadFileName`** because that's the only place where it is maintained after a file is first uploaded in Permanent. In other words permanent takes a file and then removes its extension and then what is left is called the `displayName`
```
kfogel, line 117: What is the "it" in the first sentence?

docs/collisions.md
```diff
@@ -64,6 +64,61 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five
...
+
+b) The **source string for file extensions ON FOLDERS** is the `displayName` because Permanent does not (or at least did not ) do any kind of extension proccessing on folders, hence folder extensions are always visible in their `displayName`.
```
kfogel, line 119: Just curious: Do folders also have an `uploadName`, and we're just not using it here?  Or do they not have `uploadName` in the first place?

Fenn-CS:
Thanks for the review @kfogel would update this with more details and corrections.

docs/collisions.md
Fenn-CS: Yes folders have `uploadName` however, it's not used because there's not a time we ever really need it. When a folder is first uploaded the `displayName === uploadName` unlike with files where the extension is ripped out of the `displayName`.

kfogel:
+1 on the "stop tracking implementation" commit, FWIW -- thanks, @Fenn-CS.

docs/collisions.md
```diff
@@ -64,6 +64,62 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five
 * [Extensions in folder names matter](https://apple.stackexchange.com/questions/123001/renamed-folder-becomes-a-file-with-an-extension) and should be handled :
 * Case sensitivity is taken into account. `Myfile` and `myFile` are different for example. That is useful to [ensure complete syncs](https://rclone.org/overview/#case-insensitive).
 
+#### Extensions
+
+For both files and folders the extension is any string that follows after the last dot.
```
kfogel, line 69: I was going to ask if this means `uploadFileName` or in `displayName` or somewhere else, but I see you answer that later.  You might want to put a parenthetical aside here, saying that you're going to address that question below, since it's something the reader will wonder about right away.

docs/collisions.md
```diff
@@ -64,6 +64,62 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five
...
+      </tr>
+      <tr>
+         <td>Folder.Work Files</td>
+         <td>Work Files</td>
+         <td>Extension contains spaces</td>
+      </tr>
+   </tbody>
+</table>
+
+As seen, extension can take infinite number of "shapes" BUT as long as all characters in the file name including the extension are valid file name chacters, then, **the extension is the string that follows after the last dot** (Emphasis).
```
kfogel, line 112: This raises the question of files that contain characters which are *not* valid in a filename -- but doesn't answer that question :-).  Can we ever have such characters, or do we disallow them in the first place?  If they're not allowed and can never be there, then we don't need to address them.  If we might encounter them, then that's a separate problem, and maybe not one to be documented here?  It's not really related to collision handling.

Also, in general I don't think we need all these examples.  If you just say "The extension is everything following the last dot (".", ASCII/Unicode 46) in the filename, and if the filename has no dots then it has no extension." that's clear, totally unambiguous, and it's all we need to say.  You could add something like this: "This means that an extension might contain spaces, parentheses, and other characters not traditionally thought of as extension characters.  It also means that the last dot could also be the first dot, in which case the extension is the remaining string after that first dot." if you want to make sure people understand that we really, literally mean everything after the last dot.

docs/collisions.md
```diff
@@ -64,6 +64,62 @@ For example if a folder has five 5 with the same name say **_A.txt _** the five
...
+
+
+#### Source of extensions
+
+a) The **source string for file extensions ON FILES** could be the **`displayName`** OR the **`uploadFileName`**. When a file is uploaded to Permanent if the extension is recognized, then the `displayName` is calculated as `uploadFileName - File Extension`. So, if `My File.txt` is uploaded the `displayName` is `My File` while the `uploadFileName` would be `My File.txt`. This means when Permanent recognizes the extension, it removes it (the extension) from the `displayName` and keeps track of it(the extension) via the `uploadFileName`. When Permanent does not recognize the extension, the `uploadFileName` is thesame as `displayName` **at the point of creation**. For exampl, in the case of `File My.abcxyz` both the `uploadFileName` and `displayName` would be `My File.abcxyz` and so in this case the source of the extension is the `displayName` even though the extension can be obtained from the `uploadFileName` it risk being wrong if the user ever updates the `displayName`. This is the case because an update to the `displayName` does not affect the `uploadFileName` correctly so.
```
kfogel, line 117: Best to explain what "recognized" means in this context.  And I think you're using " - " as a minus sign -- that is, subtracting the extension from the `uploadFileName`.  But it's easy to read it as a literal space-hyphen-space.  So it's better to write it out in words: "...`displayName` is calculated as `uploadFileName` minus the extension. ..."

You can avoid an awkward parenthetical by avoiding pronouns in the first place: "...This means when Permanent recognizes the extension, we don't include the extension in `displayName` but we keep track of the extension via `uploadFileName`. ..."


Typo: "thesame" should be "the same".

Another typo: "exampl".  (Suggestion: run changes through an automated spell-checker before committing!  I do that all the time with documentation commits, FWIW.)

Now, the documentation I am reading here is obsolete, right, because we're just always going to use `uploadFileName` to get the extension, because the extension is, by definition, always available there?

docs/collisions.md
kfogel: When you say "it's not used" do you mean its value is NULL?  In other words, our proposed method of just always looking at `uploadFileName` wouldn't work, because for folders there's no string there at all, and we have to look at `displayName`?

If so, our conditional should not be `if (is_file(foo)): ...`, but rather `if (uploadName is not NULL): ... ; else ...look at displayName...`, I think?

docs/collisions.md
```diff
@@ -377,51 +433,15 @@ _What each unsupported character encodes to has to ultimately be decided and the
...
-</table>
-
-
 ## Testing Plan
 
-- Test generates unique & correctly formatted `downloadName` for n colliding files in the same namespace (parent folder).
-- Test generates unique & correctly formatted `downloadName` for n colliding folders in the same namespace.
-- Test generates unique & correctly formatted `downloadName` for n colliding files & folders in the same namespace.
-- Test generates unique & correctly formatted `downloadName` for incoming files and folders holding colliding deduplication strings.
+- Test generates directory-unique & correctly formatted `downloadName` for n colliding files in the same namespace (parent folder).
```
kfogel, line 438: What's the difference between this test and the one below it?

docs/collisions.md
```diff
@@ -377,51 +433,15 @@ _What each unsupported character encodes to has to ultimately be decided and the
...
+- Test generates directory-unique & correctly formatted `downloadName` for n colliding folders in the same namespace.
+- Test generates directory-unique & correctly formatted `downloadName` for n colliding files & folders in the same namespace.
```
kfogel, line 440: What does this mean?  Does it mean "folders whose names collide with filename"?  Or does it mean "in the same namespace, we have some files that collide with other files, and some folders that collide with other folders"?

(And as per above: is a namespace the same as a folder?  We should just say folder if we mean folder.

kfogel (changes requested):

docs/collisions.md
Fenn-CS: True! Resolved.

> For both files and folders the extension is any string that follows after the last dot in the `displayName` or `uploadFileName` (see; Source of extensions section below).

docs/collisions.md
Fenn-CS: Thanks @kfogel this is really good! Done.

docs/collisions.md
Fenn-CS: Resolved!

docs/collisions.md
Fenn-CS: Yeah, namespace is folder, so I would just use it and suggested, thanks!

docs/collisions.md
Fenn-CS: Reword test to make it more explicit.

Fenn-CS:
Thanks for the review @kfogel. I have addressed all your comments in https://github.com/PermanentOrg/sftp-service/pull/23/commits/60a0f90da099d8ad88d366130264f300a5dab916 and invite you to take another holistic look at the document.

FYI : @slifty

cecilia-donnelly (approved):
Thank you, @Fenn-CS !  LGTM.

PR #27, opened 2022-07-14
Fix Got vulnerability

slifty:
This PR updates nodemon in order to address a vulnerability in one of nodemon's dependencies


Fenn-CS (approved):
Thanks @slifty and the very annoying but useful @dependabot

PR #28, opened 2022-07-14
Support loading archives from Permanent

slifty:
This PR adds support for loading archives from permanent and translating them to proper SFTP responses.

It does NOT support real authentication yet, and instead uses a temporary bearer token environment variable to handle the authentication use case.

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -1,14 +1,67 @@
+import path from 'path';
+import { v4 as uuidv4 } from 'uuid';
+import ssh2 from 'ssh2';
 import { logger } from '../logger';
+import {
+  generateFileEntry,
+  getFileType,
+  generateDefaultAttributes,
+  loadPath,
+} from '../utils';
+import type {
+  Attributes,
+  FileEntry,
+  SFTPWrapper,
+} from 'ssh2';
 
-export class SftpStreamHandler {
+const SFTP_STATUS_CODE = ssh2.utils.sftp.STATUS_CODE;
+
+const generateHandle = (): string => uuidv4();
+
+export class SftpSessionHandler {
+  private readonly authToken: string;
+
+  private readonly sftpConnection: SFTPWrapper;
+
+  private readonly openDirectories: Map<string, FileEntry[]> = new Map();
+
+  private readonly openFiles: Map<string, Buffer> = new Map();
+
+  public constructor(
+    sftpConnection: SFTPWrapper,
+    authToken: string,
+  ) {
+    this.sftpConnection = sftpConnection;
+    this.authToken = authToken;
+  }
+
   /**
    * See: SFTP events (OPEN)
    * https://github.com/mscdex/ssh2/blob/master/SFTP.md
    * Also: Opening, Creating, and Closing Files
    * https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.3
    */
-  public openHandler = (): void => {
+  public openHandler = (
+    reqId: number,
+    filename: string,
+    flags: number,
+    attrs: Attributes,
+  ): void => {
     logger.verbose('SFTP file open request (SSH_FXP_OPEN)');
+    logger.debug('Request:', {
+      reqId,
+      filename,
+      flags,
+      attrs,
+    });
+    const handle = generateHandle();
+    logger.debug(`Opening ${filename}: ${handle}`);
+    this.openFiles.set(handle, Buffer.from('content goes here'));
+    logger.debug('Response:', { reqId, handle });
+    this.sftpConnection.handle(
+      reqId,
+      Buffer.from(handle),
+    );
   };
 
   /**
@@ -57,8 +110,12 @@ export class SftpStreamHandler {
    * Also: Opening, Creating, and Closing Files
    * https://datatracker.ietf.org/doc/html/draft-ietf-secsh-filexfer-02#section-6.3
    */
-  public closeHandler = (): void => {
+  public closeHandler = (reqId: number, handle: Buffer): void => {
     logger.verbose('SFTP close file request (SSH_FXP_CLOSE)');
+    logger.debug('Request:', { reqId, handle });
+    this.openDirectories.delete(handle.toString());
+    logger.debug('Response: Status (OK)', { reqId }, SFTP_STATUS_CODE.OK);
+    this.sftpConnection.status(reqId, SFTP_STATUS_CODE.OK);
   };
 
   /**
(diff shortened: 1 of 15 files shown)


src/classes/SshSessionHandler.ts
```diff
@@ -1,36 +1,45 @@
...
-    sftp.on('STAT', sftpStreamHandler.statHandler);
-    sftp.on('REMOVE', sftpStreamHandler.removeHandler);
-    sftp.on('RMDIR', sftpStreamHandler.rmDirHandler);
-    sftp.on('REALPATH', sftpStreamHandler.realPathHandler);
-    sftp.on('READLINK', sftpStreamHandler.readLinkHandler);
-    sftp.on('SETSTAT', sftpStreamHandler.setStatHandler);
-    sftp.on('MKDIR', sftpStreamHandler.mkDirHandler);
-    sftp.on('RENAME', sftpStreamHandler.renameHandler);
-    sftp.on('SYMLINK', sftpStreamHandler.symLinkHandler);
+    const sftpStream = accept();
```
slifty, line 20: You know what -- I think all this should be `sftpSession` / `SftpSessionHandler` to mirror `SshSessionHandler`.

src/classes/SftpSessionHandler.ts
```diff
@@ -0,0 +1,388 @@
...
+import { logger } from '../logger';
+import type {
+  Attributes,
+  FileEntry,
+  SFTPWrapper,
+} from 'ssh2';
+import type { Archive } from '@permanentorg/sdk';
+
+const getFileType = (filePath: string): number => {
+  // TODO: this is not a real rule -- directories can have extensions and files don't have to have them.
```
slifty, line 15: This will end up being a lookup based on the permanent API's return results once we're loading real data.  For now we have to spoof it like this.

src/classes/SftpSessionHandler.ts
```diff
@@ -0,0 +1,388 @@
...
+  return archives.map((archive: Archive) => {
+    const archiveDirectoryPath = `/archives/${archive.name} (${archive.id})`;
+    return generateFileEntry(
+      archiveDirectoryPath,
+      generateDefaultAttributes(fs.constants.S_IFDIR),
+    );
+  });
+};
+
+const loadArchive = async (authToken: string, archiveName: string): Promise<FileEntry[]> => {
```
slifty, line 83: This is all from the original demo -- once this is merged I'm going to be swapping in the shiny new 0.2.0 archive functionality so this is all driven by the perm API / SDK

src/classes/SftpSessionHandler.ts
```diff
@@ -0,0 +1,388 @@
...
+
+const loadPath = async (requestedPath: string, authToken: string): Promise<FileEntry[]> => {
+  if (requestedPath === '/') {
+    return [
+      generateFileEntry(
+        '/archives',
+        generateDefaultAttributes(fs.constants.S_IFDIR),
+      ),
+    ];
+  } else if (requestedPath === '/archives') {
```
slifty, line 112: Maybe these should go in a named function so it's clear something like `isArchiveRootPath`

src/classes/SftpSessionHandler.ts
```diff
@@ -0,0 +1,388 @@
...
+    return loadArchives(authToken);
+  } else if (requestedPath.includes('/archives') && requestedPath.split('/').length == 3) {
```
slifty, line 114: This could be `isArchiveChildItemPath`

slifty: I'm realizing our linter is busted because it didn't yell about `==` so that's fun -- (it's the issue with airbnb rules not loading)

src/classes/SftpSessionHandler.ts
```diff
@@ -0,0 +1,388 @@
...
+    generateFileEntry(
+      `${archiveName}/Public`,
+      generateDefaultAttributes(fs.constants.S_IFDIR),
+    ),
+    generateFileEntry(
+      `${archiveName}/Apps`,
+      generateDefaultAttributes(fs.constants.S_IFDIR),
+    ),
+  ];
+};
```
cecilia-donnelly, line 102: So I thought our discussion [in the SDK](https://github.com/PermanentOrg/permanent-sdk/issues/19) led to the idea that the SDK would return the special root folders and so you wouldn't have to create them here.  Is this temporary or am I missing something?

src/classes/SftpSessionHandler.ts
cecilia-donnelly, line 114: The length here is from the place where `rclone` is initiated?  So `/archives/ArchiveName (1)/` is how you get to 3?

cecilia-donnelly (commented):
I don't have a local copy of this set up so I can't do a full test and I'm still getting my mind wrapped around this, but here are some questions/thoughts at least.

src/classes/SftpSessionHandler.ts
cecilia-donnelly: Ah you answered before I asked!

slifty: In fact, I'll mention that in a comment just because it IS so intentionally temporary.

src/classes/SftpSessionHandler.ts
```diff
@@ -0,0 +1,388 @@
...
+
+const getLongname = (
+  filename: string,
+  attributes: Attributes,
+  owner = 'nobody',
+  group = 'nogroup',
+): string => {
+  const directoryFlag = isDirectoryMode(attributes.mode) ? 'd' : '-';
+  return `${directoryFlag}rwxrwxrwx 1 ${owner} ${group} 3 Dec 8 2009 ${filename}`;
```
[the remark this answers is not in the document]
slifty: It's very specific isn't it...

No!  I'm going to replace it with a parameter (and have the parameter default to unix epoch or something more useful)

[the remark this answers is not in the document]
slifty: Actually I'm taking it out (I believe longname is pretty much... whatever we want it to be and doesn't require a date)

src/classes/SftpSessionHandler.ts
slifty: Correct!  Temporary -- this PR is built against v0.1.0 which doesn't have that functionality yet.

src/classes/SftpSessionHandler.ts
slifty: Yeah that's the idea (and again I'm gonna modify this code to be a little more clear from reading what's going on)

slifty:
I just refactored a bit to make this actually consumable (instead of having almost all utilities in the one file).

I think this is ready for another review @cecilia-donnelly

I know it's a little hard to review code that has "spoofed" values (and that you can't actually invoke without a bearer token).

Fenn-CS (commented):
First pass.

This is looking good so far. Things such as the way we determine `File` or `Directory`, auth mechanism and other things are marked as todo's and so no need to dive into them.

Fenn-CS (approved):

cecilia-donnelly (commented):
I don't think I have much to add unfortunately.  Is there a specific section you'd like me to look at in particular, @slifty ?

src/classes/SftpSessionHandler.ts
```diff
@@ -1,14 +1,67 @@
...
     logger.verbose('SFTP file open request (SSH_FXP_OPEN)');
+    logger.debug('Request:', {
+      reqId,
+      filename,
+      flags,
+      attrs,
+    });
+    const handle = generateHandle();
+    logger.debug(`Opening ${filename}: ${handle}`);
+    this.openFiles.set(handle, Buffer.from('content goes here'));
```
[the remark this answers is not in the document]
slifty: Probably! 😂   It for sure won't last in the long run.

Honestly I think that I'm gonna merge this -- this is a weird repository since there's so much R&D and demo associated with it.

slifty:
Thanks all!  I'm merging since this repository is still very much in flux and this is just one chunk of progress closer to things working.

PR #34, opened 2022-08-02
Load folders from Permanent

### LINKED ISSUE #32: Use getFolder from the permanent SDK
We want to be able to load the contents of a given folder from permanent.

This will be our first time needing to be able to convert a path to a folder ID in an archive, which is going to be interesting to figure out.

slifty:
This PR uses the Permanent SDK to load folder information from the Permanent API.

Resolves #32


Fenn-CS (approved):
This does look good.

Thought
-------------
Assuming that your intention is that files would be cached as well, what do you think about having a `pathCache` instead of `folderCache` and eventually `fileCache` (`recordCache` whatever)???

slifty:
> Assuming that your intention is that files would be cached as well, what do you think about having a pathCache instead of folderCache and eventually fileCache (recordCache whatever)???

I actually sarted with `pathCache`!  The issue is that by conflating the types of objects the cache can hold I would need to have type guards after every access point.  By having two separate caches for different types of object I get the type safety for "free".

I actually did start to add type guards for this very reason (see https://github.com/PermanentOrg/permanent-sdk/issues/55 and https://github.com/PermanentOrg/permanent-sdk/tree/55-sdk-typeguards) but then stepped back and just shifted the plan.

(Hope that all makes sense)

PR #36, opened 2022-08-04
Support file access

### LINKED ISSUE #35: Support file access (from Permanent)
We need to handle the various file OPEN / CLOSE / READ requests and also load real data provided by the permanent API (or CDN in the case of file data).

slifty:
This PR adds support for loading files from Permanent via SFTP calls.

With this functionality added, running rclone against a permanent account should successfully download all data associated with that account.

Resolves #35


Fenn-CS (approved):
LGTM.

PR #40, opened 2022-09-23
Support authentication

### LINKED ISSUE #15: Set up Authentication
We want to allow SFTP users to authenticate using a fusion token.  This is *INSTEAD* of a username and password.

Right now the [authentication logic](https://github.com/PermanentOrg/sftp-service/blob/6246c7cb1f5081cf9d999454b056d3a6552cea71/src/classes/SshConnectionHandler.ts#L13-L29) is essentially a no-op.

The handler tells the client to use password authentication, but then it accepts *any* password.  in reality we need the password to be a valid bearer token associated with a FusionAuth session.  That token needs to be validated and then saved in order to enable future API calls.  We may want to require the username to match the relevant account ID as an extra layer of auth protection, but in theory the bearer token could could serve as the sole source of authentication.

This validation *should* be run through the currently-empty [permanent SDK](https://github.com/PermanentOrg/permanent-sdk).  With that in mind, this issue should have a pretty small foot print on this particular repository once the authentication has been implemented in the SDK -- it would just call the token validation method and respond accordingly.

slifty:
This PR adds support for authentication to the sftp service (before we were using a "hard coded" authentication so a given service would only work against a single local user).

This PR also fixes some bugs that I found when utilizing the OpenSSH cli client, though I may sneak them out of this PR and merge them seperately.

Resolves #15

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -36,5 +36,7 @@ export NODE_TLS_REJECT_UNAUTHORIZED=0
 # e.g. http://local.permanet.org/api
 PERMANENT_API_BASE_PATH=${LOCAL_TEMPORARY_AUTH_TOKEN}
 
-# Temporary Authentication Token
-LOCAL_TEMPORARY_AUTH_TOKEN=${LOCAL_TEMPORARY_AUTH_TOKEN}
+# FusionAuth credentials
+# See https://fusionauth.io/docs/v1/tech/apis/api-keys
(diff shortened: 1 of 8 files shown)


Fenn-CS (approved):
LGTM, thanks!

cecilia-donnelly:
@meisekimiu even though this was approved and merged it would still be good to you to look over it so you're aware of what we're doing for auth here.

PR #43, opened 2022-10-13
Add an upload workflow

### LINKED ISSUE #42: Generate tar and upload
Permanent projects seem to have an upload / deploy workflow that is used as part of the infrastructure deployment.  For example this: https://github.com/PermanentOrg/notification-service/blob/main/.github/workflows/deploy.yml

We need that here in order to deploy with the same process.

slifty:
This PR adds the upload / bundle scripts that we use in other repositories for our deployment processes

Resolves #42


Fenn-CS (approved):
Looks okay. Thanks!

PR #49, opened 2022-12-21
Implement `MKDIR`

### LINKED ISSUE #26: Implement `MKDIR` handler
This will allow SFTP clients to create folders in their permanent archives.

slifty:
This PR adds support for folder creation by implementing the `MKDIR` sftp command handlers.

Resolves #26

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -1,6 +1,7 @@
 import path from 'path';
 import fs from 'fs';
 import {
+  createFolder,
   getArchives,
   getArchiveFolders,
   getFolder,
@@ -140,6 +141,29 @@ export class PermanentFileSystem {
       return this.loadFolderFileEntries(requestedPath);
     }
     return [];
+  }
+
+  public async makeDirectory(requestedPath: string): Promise<Folder> {
+    if (isRootPath(requestedPath)) {
+      throw new Error('You cannot create new root level folders via SFTP.');
+    }
+    if (isArchiveCataloguePath(requestedPath)) {
+      throw new Error('You cannot create new archives via SFTP.');
+    }
+    if (isArchivePath(requestedPath)) {
+      throw new Error('You cannot create new folders at the root level of an archive via SFTP.');
+    }
+    const parentPath = path.dirname(requestedPath);
+    const childName = path.basename(requestedPath);
+    const parentFolder = await this.loadFolder(parentPath);
+    this.folderCache.delete(parentPath);
+    return createFolder(
+      this.getClientConfiguration(),
+      {
+        name: childName,
+      },
+      parentFolder,
+    );
   }
 
   public async loadFile(requestedPath: string): Promise<File> {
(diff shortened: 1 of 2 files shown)


src/classes/SftpSessionHandler.ts
```diff
@@ -380,4 +382,35 @@ export class SftpSessionHandler {
...
+            attrs,
+          );
+        })
+        .catch(() => {
+          this.sftpConnection.status(reqId, SFTP_STATUS_CODE.NO_SUCH_FILE);
+        });
+    } else {
+      this.permanentFileSystem.loadDirectory(handle.toString())
+        .then(() => {
+          const attrs = generateDefaultAttributes(fileType);
```
slifty, line 404: It would be great to DRY this section somehow.

src/classes/SftpSessionHandler.ts
```diff
@@ -380,4 +382,35 @@ export class SftpSessionHandler {
   public symLinkHandler = (): void => {
     logger.verbose('SFTP create symlink request (SSH_FXP_SYMLINK)');
   };
+
+  private readonly genericStatHandler = (reqId: number, handle: Buffer): void => {
+    const fileType = getFileType(handle.toString());
+    if (fileType === fs.constants.S_IFREG) {
+      this.permanentFileSystem.loadFile(handle.toString())
```
slifty, line 389: This actually should be `loadRecord` rather than `loadFile`

src/classes/SftpSessionHandler.ts
```diff
@@ -380,4 +382,35 @@ export class SftpSessionHandler {
...
+        .then(() => {
+          const attrs = generateDefaultAttributes(fileType);
```
slifty, line 391: We should populate the attributes correctly.

src/classes/PermanentFileSystem.ts
```diff
@@ -142,6 +143,29 @@ export class PermanentFileSystem {
...
+    if (isArchiveCataloguePath(requestedPath)) {
+      throw new Error('You cannot create new archives via SFTP.');
+    }
+    if (isArchivePath(requestedPath)) {
+      throw new Error('You cannot create new folders at the root level of an archive via SFTP.');
+    }
+    const parentPath = path.dirname(requestedPath);
+    const childName = path.basename(requestedPath);
+    const parentFolder = await this.loadFolder(parentPath);
+    this.folderCache.delete(parentPath);
```
Fenn-CS, line 159: I guess the reason for this, is to only cache paths that the user has explicitly loaded not, just ones that were loaded due to their own creation?

slifty: The cache needs to be dirtied because it is no longer accurate -- the cached version of the folder doesn't know about the newly created folder.

I am going to make an issue about this but I fear that the way rclone parallelizes it will result in multiple connections (and therefore multiple caches).

slifty: Issue created: https://github.com/PermanentOrg/sftp-service/issues/76

Fenn-CS: +1

PR #51, opened 2023-01-24
How to run rclone

Fenn-CS:
This commit adds instructions on how to run `rclone` against the various Permanent instances.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,50 @@ or, if you want to start the development environment
 npm run dev
 ```
 
+## Running [rclone](https://rclone.org) against Permanent.org instances
+
+- Install rclone, using [official install guide](https://rclone.org/install/)
+
+### Setting up an rclone remote
+
+- `rclone config`
+- Select option n for new and enter a remote name such as `permanent-prod` or `permanent-dev`.
+- Set `SFTP` as storage protocol by by selecting from displayed menu, after name prompt.
+- Set an ssh host, the host format is `sftp.${environment}.permanent.org` where `${environment}` is either `dev` or `staging` for the dev and staging environments respectively.
+    - For example `sftp.dev.permanent.org` (for dev).
+    - Prod has ssh host simply as `sftp.permanent.org` (without specifying environment).
+- Set your permanent username once prompted for example `rclone.joy@example.com`.
+- Set `SFTP` port as `22` which is the default.
+- Set `ssh` password by selecting the "Yes, type in my own password" (`y`) option, when the password options show up.
+- Set your remote password which is the password to your permanent account (in this case `rclone.joy@example.com`).
+- Set every other option from this point to the end, with its default value by simply pressing enter.
+
+At the end of the configuration you should have something like this:
+
+```
+Configuration complete.
+Options:
+- type: sftp
+- host: sftp.permanent.org
+- user: rclone.joy@example.com
+- pass: *** ENCRYPTED ***
+Keep this "permanent-prod" remote?
+y) Yes this is OK (default)
+e) Edit this remote
+d) Delete this remote
+```
+If everything looks okay, you want to select `y` (yes) to complete the remote setup. After this, you should be able to quit the configuration interface.
+
+### Running rclone clone against remote
+
+Run the following command sync down from the newly set remote:
+
+    - `rclone copy permanent-prod:/ ~/permanent-sync-folder` (where `permanent-sync-folder` is any local destination for the syncing)
+
+### Troubleshooting rclone
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -35,6 +35,44 @@ or, if you want to start the development environment
...
+## Running [rclone](https://rclone.org) against Permanent.org instances
+
+- Install rclone, using [official install guide](https://rclone.org/install/)
+
+### Setting up an rclone remote
+
+- `rclone config`
+- Select option n for new and enter a remote name such as `permanent-prod` or `permanent-dev`.
+- Set `SFTP` as storage protocol by by selecting from displayed menu, after name prompt.
+- Set an `ssh` host, the host format is `sftp.${environment}.permanent.org` for example `sftp.dev.permanent.org` for `dev` and `staging`. The prod environment is an exception and simply has host as `sftp.permanent.org`.
```
cecilia-donnelly, line 47: Can you wrap this line? Also is it really `sftp.dev.permanent.org` for both dev and staging?

README.md
```diff
@@ -35,6 +35,44 @@ or, if you want to start the development environment
...
+- Set you permanent username once prompted for example `rlcone.joy@gmail.com`.
```
cecilia-donnelly, line 48: It's not a big deal, but having a misspelling of rclone in this example email is a bit confusing. It's also better practice to use example.com for these, I think, since this might be a real email address.

README.md
```diff
@@ -35,6 +35,44 @@ or, if you want to start the development environment
...
+- Set every other option from this point to the end, with its default value by simply pressing enter.
+
+At the end of the configuration you should have something like this:
+
+```
+Configuration complete.
+Options:
+- type: sftp
+- host: sftp.permanent.org
+- user: nfebe@opentechstrategies.com
```
cecilia-donnelly, line 61: Better to use the same example email here that you use above.

README.md
Fenn-CS: Thanks @cecilia-donnelly, I would rephrase that... What I really mean is that, the `environment` part of the `ssh` host is dynamic for `dev` and `staging`.

So the user, can either use `sftp.dev.permanent.org` for dev or `sftp.staging.permanent.org`...

What about 

> Set an `ssh` host, the host format is `sftp.${environment}.permanent.org` where `${environment}` is either `dev` or `staging` for the dev and staging environments respectively.  For example `sftp.dev.permanent.org.  Prod simply has ssh host as `sftp.permanent.org`.

README.md
```diff
@@ -35,6 +35,44 @@ or, if you want to start the development environment
...
+### Setting up an rclone remote
+
+- `rclone config`
+- Select option n for new and enter a remote name such as `permanent-prod` or `permanent-dev`.
+- Set `SFTP` as storage protocol by by selecting from displayed menu, after name prompt.
+- Set an `ssh` host, the host format is `sftp.${environment}.permanent.org` for example `sftp.dev.permanent.org` for `dev` and `staging`. The prod environment is an exception and simply has host as `sftp.permanent.org`.
+- Set you permanent username once prompted for example `rlcone.joy@gmail.com`.
+- Set `SFTP` port as `22` which is the default.
+- Set `ssh` password by selecting the "Yes, type in my own password" (`y`) option, when the password options show up.
+- Set your remote password which is the password to your permanent account (in this case `rlcone.joy@gmail.com`).
```
cecilia-donnelly, line 51: Thanks for describing this. It is a little confusing!

README.md
```diff
@@ -35,6 +35,44 @@ or, if you want to start the development environment
...
+- type: sftp
+- host: sftp.permanent.org
+- user: nfebe@opentechstrategies.com
+- pass: *** ENCRYPTED ***
+Keep this "permanent-prod" remote?
+y) Yes this is OK (default)
+e) Edit this remote
+d) Delete this remote
+```
+If everything looks okay, you want to select okay to complete the remote setup.
```
cecilia-donnelly, line 68: I'd say "y" instead of "okay" here and then make clear that the user can quit the config afterward.

cecilia-donnelly (changes requested):
Thanks for writing this up, @Fenn-CS !

cecilia-donnelly (approved):
Thank you for writing this up, @Fenn-CS !

Fenn-CS:
Thanks @cecilia-donnelly buhh sorry I pushed another small improvement with wrapping, FYI.

slifty (approved):
Thanks for writing this up!  Let's merge it in (it can be iterated on as we have more instructions)

PR #53, opened 2023-01-26
Update instructions for using rclone w/ Permanent

kfogel:


README.md
```diff
@@ -37,43 +37,82 @@ npm run dev
...
+Run **`rclone config`** and answer the questions it asks:
 
-At the end of the configuration you should have something like this:
+- Select option **`n`** (for "new") and make a remote name (e.g., "permanent").
+
+- At the **`Storage>`** prompt, set the storage protocol to SFTP, either by typing "sftp" at the prompt or by selecting the corresponding number from the displayed menu (the numbered entry may say "SSH/SFTP").
+
+- At the **`host>`** prompt, set the host to **`sftp.permanent.org`**.
+
+  If you want to run rclone against the dev or staging environment, then use `sftp.dev.permanent.org` or `sftp.staging.permanent.org` instead.  (You can also say `sftp.prod.permanent.org`, but it means the same thing as just `sftp.permanent.org`.)
```
Fenn-CS, line 53: > (You can also say `sftp.prod.permanent.org`, but it means the same thing as just `sftp.permanent.org`.)

I wonder if this is true.

kfogel: Well, I'm merely repeating a claim that was in the original.

kfogel: > I wonder if this is true.

It is not true -- I misread some material about dev and staging in the original text, and mistakenly applied it to prod as well.  Thanks for catching this!  I've pushed the fix (commit e1e92104712) to this branch.

Fenn-CS (approved):
Thanks for the upgrade @kfogel

kfogel:
Merged!  Thanks, @Fenn-CS.

PR #58, opened 2023-02-01
Refactor folder opening

### LINKED ISSUE #57: If a folder has a flawed file the whole folder fails to load
When opening a folder the SFTP service process all the immediate children responses (in order to generate file stats to return) before continuing.

If any of those children are flawed (for instance, if they don't have an original file), then the OPEN will fail.

In an ideal world the folder would simply skip flawed files as if they didn't exist, rather than having the whole command fail.

slifty:
This PR improves folder opening logic to skip any records that don't have original files (rather than simply erroring out).

Resolves #57

### DIFF
--- a/src/utils/getOriginalFileForRecord.ts
+++ b/src/utils/getOriginalFileForRecord.ts
@@ -1,0 +1,15 @@
+import { DerivativeType } from '@permanentorg/sdk';
+import type {
+  File,
+  Record,
+} from '@permanentorg/sdk';
+
+export const getOriginalFileForRecord = (record: Record): File => {
+  const originalFile = record.files.find(
+    (file) => file.derivativeType === DerivativeType.Original,
+  );
+  if (!originalFile) {
+    throw Error(`Permanent does not have an original file for record ${record.id}`);
+  }
+  return originalFile;
+};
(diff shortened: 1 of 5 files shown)


src/utils/getOriginalFileForRecord.ts
```diff
@@ -0,0 +1,15 @@
+import { DerivativeType } from '@permanentorg/sdk';
+import type {
+  File,
+  Record,
```
Fenn-CS, line 4: Unrelated reminder : We should update the `permanent-sdk` being used. This is now `ArchiveRecord`.

slifty: Yes indeed! -- that won't happen before the demo, but it will happen (along with a lot of other related changes) as part of the next PR (there's a working branch with those changes already)

Fenn-CS (approved):
+ 1 on moving `generateFileEntries` for `records` and `folders` to their own files.
+ 1 on adapting the initial file entry map, with a reduce and a try that would ultimately remove file without the original files as expected.

src/utils/getOriginalFileForRecord.ts
Fenn-CS: Okay!

PR #59, opened 2023-02-01
In usage instructions, clarify SSH's role earlier

kfogel:
The `rclone config` prompts mention "SSH" in a possibly-confusing way when asking for username.  Therefore, I've moved the part of our instructions that clarifies why rclone is mentioning "SSH" at all to a bit earlier, to line up with that mention.  (It could also go where the "host>" is prompted, right before the username prompt, but I think confusion is most likely to happen with username because people have a pre-existing idea of what their SSH username is.)

This will help users who have experience with SSH better understand what's going on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,11 @@ Run **`rclone config`** and answer the questions it asks:
 
 - At the **`user>`** prompt, enter your Permanent username, which is the email address associated with your account.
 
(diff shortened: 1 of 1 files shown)


Fenn-CS (approved):
Thanks Karl!

PR #62, opened 2023-02-01
Use fileSystemCompatibleName for items

### LINKED ISSUE #60: Use fileSystemCompatibleName for folders
Right now we use the folder display name (since that is the only thing exposed by v0.3.0 of the SDK).  This means folders with the same display name result in duplicate folder entries.

0.4.0 of the SDK exposes the deduplicated version of each folder name, and so it's time to update!

slifty:
This PR updates the SFTP service to use the latest Permanent SDK, and in particular stops using the Folder "name" property (which does not guarantee filesystem safety), instead using `fileSystemCompatibleName`.

It also updates `fileName` on Records to the new attribute name.

Resolves #60

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@fusionauth/typescript-client": "^1.38.0",
-        "@permanentorg/sdk": "^0.3.0",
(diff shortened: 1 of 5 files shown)


Fenn-CS (approved):
Thanks @slifty, LGTM.

PR #66, opened 2023-02-02
Fix record lookup

### LINKED ISSUE #65: Files are no longer being returned
Oh dear -- this was just a flat out bug (and is why we need integration tests) introduced by #58.

Folders returned by the permanent SDK return shallow Records -- in particular, the records don't have any populated files (in fact, our types should almost certainly signal this, somehow, by indicating that `FileVOs` can be null OR we should make sure they're populated, or something other than what's happening.  That's another issue though).

The point is: we can't simply iterate through the Records in a folder -- we need to actually / explicitly load the files via the permanent file system before generating the file metadata.

slifty:
This PR adds steps to populate the files of `Record` objects before generating file entries when opening a folder. This in turn means that files will be reported when listing directory contents over SFTP.

Resolves #65

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -147,6 +147,40 @@ export class PermanentFileSystem {
     return populatedRecord;
   }
 
+  private async loadRecords(requestedPaths: string[]): Promise<Record[]> {
+    // See https://github.com/PermanentOrg/permanent-sdk/issues/73
+    // This implementation is intentionally sequential even though using
+    // requestedPaths.map would allow us to run all of the lookups in parallel.
+    // This decision is because I'm worried about crushing the Permanent server
+    // if a given folder had a huge number of records.
+    return requestedPaths.reduce<Promise<Record[]>>(
+      async (
+        recordsPromise,
+        requestedPath,
+      ) => {
+        const records = await recordsPromise;
+        return [
+          ...records,
+          await this.loadRecord(requestedPath),
+        ];
+      },
+      Promise.resolve([]),
+    );
+  }
+
+  private async loadDeepRecords(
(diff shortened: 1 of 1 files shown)


src/classes/PermanentFileSystem.ts
```diff
@@ -147,6 +147,40 @@ export class PermanentFileSystem {
     return populatedRecord;
   }
 
+  private async loadRecords(requestedPaths: string[]): Promise<Record[]> {
+    // See https://github.com/PermanentOrg/permanent-sdk/issues/73
+    // This implementation is intentionally sequential even though using
+    // requestePaths.map would allow us to run all of the lookups in parallel.
```
cecilia-donnelly, line 153: (Tiny nitpick, not a blocker) Typo, should be `requestedPaths`

cecilia-donnelly (approved):
This is sort of a fake approval. This all makes sense to me but (1) I haven't really been following the progress of this codebase so I don't know if it's ignoring some major other piece and (2) I don't have a local copy of this built so I haven't tested.

I am approving in the interest of getting something deployed so we can see what happens!

src/classes/PermanentFileSystem.ts
kfogel: Dang -- good eye, @cecilia-donnelly.

kfogel (commented):
Obviously I'm not expert enough in this code to do a proper review.  But the change had the shape I was expecting -- load shallow records, then map over them to en-deepen them -- and nothing jumps out as obviously wrong.

I wouldn't feel right saying "approve", given my limitations as a reviewer, but LGTMFWIW.

slifty:
OK thanks all!  We can always patch later if needed of course.

PR #67, opened 2023-02-02
Generate correct file attributes

### LINKED ISSUE #64: Implement stats properly
fstat / lstat etc right now return responses that don't include the proper file sizes (which may actually be part of the issue with #45).

This results in sftp client reporting issues, since rclone doesn't know how big the file transfer is going to be.

This is already implemented in a branch as part of another feature (folder creation) but it should be its own effort, as it effects other aspects of SFTP.

slifty:
This PR updates the various FSTAT handlers to actually leverage Permanent data when generating a response.

Note that this branch is based off of code included in #66, and so this PR shouldn't be reviewed until after that PR is merged and this gets rebased again.

Resolves #64

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -7,6 +7,8 @@ import {
   getRecord,
 } from '@permanentorg/sdk';
 import {
+  generateAttributesForFile,
+  generateAttributesForFolder,
   generateDefaultAttributes,
   generateFileEntry,
   generateFileEntriesForRecords,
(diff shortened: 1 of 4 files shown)


cecilia-donnelly (commented):
I'm even more in the woods on this one than #66 so I'll leave it at that, but I have read it and get the gist if you need/want me to press :+1: tomorrow morning, @slifty !

kfogel (commented):
I also looked, but like Cecilia am even more out of my depth here than I was in #66 (which I see this PR subsumes).  Uh... <weak grin>... looks good?  At least doesn't look bad?  Looks like code?  Has a convincing monospace font?

slifty:
Makes sense -- I am sorry it isn't in two commits (one to DRY the original / bad implementation, the other to improve that bad implementation).  Were it not for the time pressure I would have done that :(

I've rebased at least.

@cecilia-donnelly This should be merged in before you deploy tomorrow; depending on what time that is if I'm around I'd be glad to do a quick pair to walk you through it (unless of course @Fenn-CS has a chance to review before either of us get to it!)

At any rate: whoever approves please feel free to merge it instead of waiting for me to push that button!

Fenn-CS (approved):
Looks great, thanks!

PR #68, opened 2023-02-02
Ensure unique file entries

### LINKED ISSUE #56: Filter duplicate results from the server
Right now there is a known issue where the back end will sometimes return duplicate results.

Ideally we would be able to trust back end results -- e.g. this is a bug with the back end -- but for the short term we can defensively de-duplicate the responses from the permanent SDK.

slifty:
This PR de-duplicates file entries generated from the Permanent API response when loading a folder.

Resolves #56

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -7,6 +7,7 @@ import {
   getRecord,
 } from '@permanentorg/sdk';
 import {
+  deduplicateFileEntries,
   generateAttributesForFile,
   generateAttributesForFolder,
   generateDefaultAttributes,
@@ -293,9 +294,9 @@ export class PermanentFileSystem {
         requestedPath,
       ),
     );
-    return [
+    return deduplicateFileEntries([
       ...folderFileEntities,
       ...recordFileEntities,
-    ];
+    ]);
   }
 }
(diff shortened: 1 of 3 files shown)


src/classes/PermanentFileSystem.ts
```diff
@@ -54,6 +54,20 @@ const isItemPath = (requestedPath: string): boolean => (
   && requestedPath.split('/').length > 3
 );
 
+// TODO: this exists defensively, and maybe should continue to exist defensively,
+// but in theory the backend will stop returning duplicates at some point.
+// See https://github.com/PermanentOrg/back-end/issues/396
+const deduplicateFileEntries = (fileEntries: FileEntry[]): FileEntry[] => (
+  fileEntries.filter(
+    (filterFileEntry, index, self) => {
```
slifty, line 62: This function is more complex than I would want it to be.

Given more time I would prefer it:

1. Be in a utility file so it could be unit tested (eventually)
2. Have these sub-functions as non-exported named helper functions (e.g. `fileNameFilter` and `getFirstIndexByFilename`) to break up the nested-and-kinda-complicated array operations.

Maybe I'll do that despite the time ;)

Fenn-CS: Since you are here already....

`duplicateFileEntries` seem to suggest that the opposite of what that function is doing. Something like `removeDuplicateCopyFileEntries` or similar?

Fenn-CS: It's `deduplicate`, all good on that, sorry for the dust!

Fenn-CS (approved):
Thanks for the hedge.

PR #71, opened 2023-02-02
Add more info about using rclone with Permanent

kfogel:
In the "Troubleshooting rclone" section, add some information about how to include empty directories in 'rclone copy', link to the rclone option documentation online, and edit the existing item about MFA.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,15 @@ rclone copy permanent:/ ./my-permanent-data
 
 ### Troubleshooting rclone
 
-- MFA Support : `rclone` does not support MFA. Hence, you should disable MFA in your permanent account if you have it enabled otherwise rclone would fail.`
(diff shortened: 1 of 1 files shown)


Fenn-CS (approved):
This looks good.

_However the flag may be indicated at the end of the entire command, but it does not hurt to tell the user a better place to put it_

kfogel:
Thanks, @Fenn-CS !  I'll let users figure out alternate flag placement themselves; in these instructions I think we just want to tell them one consistent way to do it.

kfogel:
@Fenn-CS Ah, wait, I'm sorry, I read your suggestion more carefully now (I was in a meeting before) and I see what you mean.

PR #73 makes the fix you had in mind, I think.

PR #72, opened 2023-02-02
Standardize log levels

### LINKED ISSUE #70: Use better logging levels
Right now we have been using `debug` for a wide range of log messages, including things that would make sense as `verbose`  (e.g. higher level request / response logs) and things that should be `silly` (e.g. raw data being moved).

Let's fix that so that we can have more control over log levels in different environments such as production

slifty:
This PR standardizes log levels to make it easier for us to control log levels appropriately in various environments.

as a reminder the log levels are:

```
    error
    warn
    info
    verbose
    debug
    silly
```

Resolves #70

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -47,20 +47,22 @@ export class SftpSessionHandler {
     flags: number,
     attrs: Attributes,
   ): void => {
-    logger.verbose('SFTP file open request (SSH_FXP_OPEN)');
-    logger.debug('Request:', {
-      reqId,
-      filename,
-      flags,
-      attrs,
-    });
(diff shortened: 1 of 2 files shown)


src/classes/SftpSessionHandler.ts
```diff
@@ -114,7 +133,8 @@ export class SftpSessionHandler {
     })
       .then(async (response) => {
         const data = await response.buffer();
-        logger.debug('Response: Data', { reqId, data });
+        logger.verbose('Response: Data', { reqId, data });
```
slifty, line 136: Whoops!  the whole point is NOT posting data in verbose mode, need to remove this

liam-lloyd (approved):

PR #73, opened 2023-02-03
Don't wrongly imply strict flag positioning

kfogel:
This takes @Fenn-CS's suggestion from his review of PR #71.


Fenn-CS (approved):
Looks great! Thanks

PR #74, opened 2023-02-08
Fix rclone hang

### LINKED ISSUE #45: rclone hangs after completion
I've opened an [issue on the rclone forums](https://forum.rclone.org/t/rclone-does-not-end-after-finishing-an-sftp-copy/34004/2) about this to get insight there, but after completing a `copy` command rclone seems to sit there with open processes.  I'm betting there's something about our protocol implementation that is causing this issue.  Maybe we don't terminate a given connection properly somehow after a file is completely copied.

slifty:
This PR adds some fairly gnarly code to accommodate a bug in the upstream library that handles implementation of the ssh / sftp protocol's more inner workings.

There is [already an upstream patch](https://github.com/mscdex/ssh2/pull/1111) which unfortunately has not been merged yet.  Once that is merged we can delete this local patch.

Resolves #45

### DIFF
--- a/src/classes/SshConnectionHandler.ts
+++ b/src/classes/SshConnectionHandler.ts
@@ -105,9 +105,13 @@ export class SshConnectionHandler {
   ): void => {
     logger.verbose('SSH request for a new session');
     const session = accept();
-    const sessionHandler = new SshSessionHandler(this.authSession?.authToken ?? '');
+    const sessionHandler = new SshSessionHandler(
+      session,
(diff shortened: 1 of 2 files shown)


cecilia-donnelly (approved):
Thank you, @slifty! This all makes sense to me and I appreciate the detailed comment.

kfogel (approved):
Nicely done.  This change is about as clean as it possibly can be, and it explains itself well.

PR #78, opened 2023-02-15
Migrate to 0.5.0 of permanent SDK

### LINKED ISSUE #77: Update to @permanentorg/sdk v0.5.0
The latest SDK has support for record creation but it also has a few breaking changes to existing functionality (mostly around the rename of `Record` to `ArchiveRecord`).  I'd like to do that migration in order to allow file creation support.

slifty:
This PR updates our SDK version to 0.5.0 and makes the necessary changes to our code to support the SDKs new names.

Resolves #77


src/utils/generateFileEntriesForArchiveRecords.ts
```diff
@@ -0,0 +1,25 @@
+import { getOriginalFileForArchiveRecord } from './getOriginalFileForArchiveRecord';
```
slifty, line 1: Unfortunately I couldn't track this as a "move" in git because too much of the content changed alongside it.

Nothing about the implementation changed, just the variable names / called function names.

src/utils/getOriginalFileForArchiveRecord.ts
```diff
@@ -0,0 +1,15 @@
+import { DerivativeType } from '@permanentorg/sdk';
```
slifty, line 1: Same situation here -- this is a "new file" because so many lines changed, but the only actual change should be variable names.

cecilia-donnelly (approved):
LGTM though I didn't test it.

PR #86, opened 2023-02-24
Share filesystems per user

### LINKED ISSUE #76: Cache per user rather than per connection
The sftp service cache was written in in a way that assumed exactly one active sftp connection per rclone "session", which is an invalid assumption and can lead to correctness issues when there is uploading happening.

**Short summary:** When there are multiple simultaneous connections (e.g., the user just runs two rclones at the same time against the same archive, or maybe rclone itself decides to use multiple simultaneous connections as some kind of network optimization), and files/folders in Permanent are being written to, then the cached versions of those files / folders on other connections will not be updated and hilarity will ensue.  (In addition to the correctness issue, this is also an efficiency loss -- N-1 connections in the same session don't get to take advantage of the caching already done by any given connection in that session.)

**Longer explanation:** Permanent has no API that converts a path to an item ID.  If the SFTP Service is asked to open a certain path, then it has to walk back up the path and build the virtual filesystem from the top down, and only then can it know what item ID to open -- and each component along that path is an API call.  Therefore, as an optimization, we cache those responses as we go, so that we don't have to make those API calls every time: we're usually walking paths we've walked before, so let's use the answer we remembered last time.  Well, that's great for reading, but we have a cache invalidation problem if we're writing.  And we do invalidate the cache!  But we can only do so in the connection that's doing the write -- current implementation has one cache per connection.  What if the user is running multiple SFTP connections?  Or what if rclone itself decides to spin up multiple connections, the way web browsers do?

**Solution:** I believe this can be resolved by having a cache manager that provides a single cache for each authenticated user rather than on a per-connection basis.  This needs to be done thoughtfully in order to avoid risks of caches lasting longer than they ought to, or having data leak between users.  Remember that it's okay for a cache to be cleaned up too soon -- all that means is that we get a bit less efficient, because we have to make some API calls that could have been avoided if the cache were available.  On the other hand, it's *not* okay to fail to clean up caches, even occasionally: that would cause a slow-but-unbounded memory leak.

(Editorial note: This initial description was expanded by Karl, with Dan's permission, after our 2023-02-23 regular dev meeting.)

slifty:
This PR creates a "manager" layer that will provide the same filesystem for any connection that has been established on behalf of a given user.  It will also ensure that those filesystems are deleted after 5 minutes of inactivity.

Resolves #76

### DIFF
--- a/src/classes/AuthenticationSession.ts
+++ b/src/classes/AuthenticationSession.ts
@@ -15,7 +15,7 @@ enum FusionAuthStatusCode {
 export class AuthenticationSession {
   public authToken = '';
 
-  private readonly authContext;
+  public readonly authContext;
 
   private readonly fusionAuthClient;
 
--- a/src/classes/PermanentFileSystemManager.ts
+++ b/src/classes/PermanentFileSystemManager.ts
@@ -1,0 +1,51 @@
+import { logger } from '../logger';
+import { PermanentFileSystem } from './PermanentFileSystem';
+
+export class PermanentFileSystemManager {
+  private readonly permanentFileSystems = new Map<string, PermanentFileSystem>();
+
+  private readonly deletionTimeouts = new Map<string, NodeJS.Timeout>();
+
+  public getCurrentPermanentFileSystemForUser(
+    user: string,
+    authToken: string,
+  ): PermanentFileSystem {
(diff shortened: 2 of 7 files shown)


Fenn-CS (approved):
Thanks @slifty, code looks okay to me! 

Not manually tested though.

PR #87, opened 2023-02-28
Remove arrow functions for class methods

### LINKED ISSUE #85: Class methods should be functions (not arrow functions)
There is some backstory to this issue which I can dig up for those curious, but ultimately I think we concluded that class methods should be `functions` (since otherwise it makes object instantiation create new copies of each function).  There are other reasons, but the biggest is just we should stick with the decision from a consistency PoV.

slifty:
This PR updates some of our class functions so they are more consistent in NOT using arrow functions.

The functionality should all be the same, though this does impact some inner workings of how the functions themselves are created and called.

Resolves #85


Fenn-CS (approved):
Apart from the scope binding-syntax 😬 this also keeps the classes cleaner with no `=` and `=>`.

Thanks @slifty

PR #89, opened 2023-02-28
Detect and handle AWS status codes

### LINKED ISSUE #82: rclone reports unequal/incompatible size files expected versus sent.
Backend (Permanent SDK) sometimes returns a file size that does not match with what rclone expects or calculates. This results in an ERROR where rlcone warns about a corrupted file.

```
2023-02-21 20:45:15 ERROR : archives/REDACTED/My Files/Why Startups Fail - DRAFT.docx: corrupted on transfer: sizes differ 13533 vs 13608
2023-02-21 20:48:10 ERROR : archives/REDACTED/My Files/Me 2015.jpg: corrupted on transfer: sizes differ 23445 vs 23571
```

This error is only reported on the rclone side.

To investigate:

- We need to pick one of the causing files, like `Me 2015.jpg`
- Get it's file size
- Figure whether Permanent gets's the file size right and if not WHY
- What is sent to the SFTP-service
- This might translate to an issue in the [`backend`](https://github.com/PermanentOrg/back-end) but noting it here for now.

### LINKED ISSUE #88: If Permanent returns an invalid download URL our service returns invalid data to the client
While debugging #82 @Fenn-CS discovered that the files in question were populated with an AWS access error message (and indeed, when trying to download the files via the permanent frontend the download URL resolved to an AWS access error XML string).

This means that the permanent download URL is incorrect, which the SFTP service cannot fix.

What we can do, however, is detect when the server response is not a 200 http status code and pass along the resulting error to the rclone client (e.g. in this case it would be an access error SFTP status).

slifty:
This PR checks the AWS response code and only returns data if the response code indicates success.

Resolves #88 
Resolves #82

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -155,13 +155,43 @@ export class SftpSessionHandler {
       },
     })
       .then(async (response) => {
-        const data = await response.buffer();
-        logger.verbose('Response: Data', { reqId });
-        logger.silly('Sent data...', { data });
-        this.sftpConnection.data(
-          reqId,
-          data,
+        logger.silly(
+          'Fetch response status code...',
+          { statusCode: response.status },
         );
+        switch (response.status) {
+          case 206:
+          case 200: {
+            const data = await response.buffer();
+            logger.verbose('Response: Data', { reqId });
+            logger.silly('Sent data...', { data });
+            this.sftpConnection.data(
+              reqId,
+              data,
+            );
+            break;
+          }
+          case 403:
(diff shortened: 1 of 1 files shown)


Fenn-CS (approved):
My assumption is if the user, reaches this point, they are already authenticated and so we can't have a `401` otherwise everything looks okay.

slifty:
Yes -- I don't think AWS will return a 401 for this *but* if they do then it will still fall down to a more generic `FAILURE` result (which I think is OK because that's a particularly unexpected outcome for the reasons you mention)

PR #96, opened 2023-03-02
Use slug instead of archive ID

### LINKED ISSUE #91: Use Archive Number instead of Archive ID
This issue might better belong in the [SDK](https://github.com/PermanentOrg/permanent-sdk/) or [back-end](https://github.com/PermanentOrg/back-end/) repository.  If so, this ticket can become a pointer to the underlying ticket over there.  But right now I just need to get this documented, because it's one step in a towering and unexpected yak shave that I need to finish, so here's what's going on:

Using SFTP to fetch a single specific archive down, or to upload up to a specific archive, requires knowing the internal archive ID number.  For example, in a command like this...

```
rclone copy -v -P --create-empty-src-dirs --size-only --sftp-set-modtime=false \
       "permanent-prod:/archives/kfogel OTS test (13539)/My Files/" ./my-permanent-data
```

...that's the "13539" in parentheses.

The problem is that there is no way for the user to get that number from the web UI, as [@cecilia-donnelly and I were just discussing in Slack](https://permanent-org.slack.com/archives/C01ARGVA849/p1677690440272029).

Now, you can get the _public_ ID by going to the archive's [profile page](https://www.permanent.org/app/(public//dialog:profile)) while logged in and clicking the "View On Web" button in the upper right -- then the URL will switch to something like "https://www.permanent.org/p/archive/09lg-0000/profile" and you can see the "09lg-0000" in the URL.

But that is not the number we're looking for.  We're looking for something like "13539", and currently there's no way to discover it except by downloading _all_ of your archives via rclone (or whatever SFTP client) and then looking in the downloaded folder names for numbers in parentheses.

Assigning this to @cecilia-donnelly for now, because she's in the best position to decide how to / who should handle this.

slifty:
This PR changes the value used in archive folder paths from the archive's ID to the archive's slug

Resolves #91

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "AGPL-3.0",
       "dependencies": {
         "@fusionauth/typescript-client": "^1.38.0",
-        "@permanentorg/sdk": "^0.5.1",
+        "@permanentorg/sdk": "^0.5.2",
         "@sentry/node": "^6.16.1",
         "dotenv": "^10.0.0",
         "logform": "^2.3.2",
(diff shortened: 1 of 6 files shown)


Fenn-CS (approved):
Everything looks okay! Thanks @slifty

Issue #100, opened 2023-03-03
Should we have our root level folder be the username?


slifty:
Right now when you rclone down we have a `/archives` folder; this could be more meaningful as `/[username]`.

On the other hand, maybe it wouldn't be, and maybe it would just come off as confusing to have a folder like `/blah@foo.bar`.

PR #101, opened 2023-03-03
Fix mtime dates

### LINKED ISSUE #98: Make sure directory times mirror the displayDate of a folder
When running `rclone lsd` we found that the unix epoch is being used as the date linked to files.  This is not correct of course!

slifty:
This PR fixes a few bugs related to the reported mtimes for files and folders.  In particular it now properly uses seconds (instead of ms) when reporting the dates.  It also uses the archive data for archive folders.

It does not address the `/archives` folder, which does not actually exist in permanent and should likely be linked to the user's account creation date, but I'd like to address that as part of #100 

Note this can be tested locally using the `rclone lsjson` command which will show the mtimes associated with various items in a path.

Resolves #98

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -10,6 +10,7 @@ import {
 } from '@permanentorg/sdk';
 import {
   deduplicateFileEntries,
+  generateAttributesForArchive,
   generateAttributesForFile,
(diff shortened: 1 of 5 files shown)


Fenn-CS (approved):
Divisions by 1000 seem more powerful that we thought, lol.

PR #103, opened 2023-03-06
Return more specific statuses

### LINKED ISSUE #97: Return a more specific error code for archive records without populated files
When a new file is created it can take a little while for permanent to fully process the file.

While processing is still happening the archive record is still a valid object, but its original file has no size or downloadUrl.  It could take an arbitrary amount of time for these files to resolve.  During that time, the files are unreadable.  `READ` attempts will fail, and STAT data (e.g. size) will be incorrect.

I think the "real filesystem" equivalent of this situation might be that the file is unreadable (even if only temporarily).  Our service should simulate that behavior and return the proper error codes so it isn't interpreted as a "failure".  We also should not be giving a `WARN` in our logs in this circumstance, as it is known and expected behavior.

I believe this change will also address some of the issues raised in #94

slifty:
This PR returns some more specific statuses than the generic `FAILURE` for a few specific cases when opening files in various states.

Resolves #97


slifty:
Oh dear!  So this actually doesn't work because the ssh2 library doesn't recognize these statuses.

You might say: but Dan, you're using statuses defined in the ssh2 types!  To which I'll say...  "whoops, I added those to the upstream type library not realizing the upstream implementation library would throw a validation error".

I've opened a PR which I hope will be accepted and merged: https://github.com/mscdex/ssh2/pull/1279/files

In the mean time this PR should remain open.

Fenn-CS (approved):
Looks good!

Thanks @slifty

Issue #102, opened 2023-03-03
Let's be smarter with caching


slifty:
Caching is complicated!

I implemented a filesystem cache to help address the fact that we have to crawl every step of a given path for any given SFTP call.  This is fine, but the resulting implementation is a little too aggressive.

I'd like to propose the following change to the cache:

1. The cache is never used for direct queries -- that is, if opening a directory the cache will *always* be ignored for the final directory call.

2. Any time a cached folder claims to not contain a given item, we try re-loading that cached folder.

There may be some other ideas as well, but this should help fix a large number of our current cache-related bugs.

slifty:
I think caching is now about as smart as I was intending when I opened this issue... so I'll close it, knowing this probably won't be the last cache related issue...

PR #104, opened 2023-03-07
Add cache bypasses

slifty:
This PR adds some limits to when the cache is used in an attempt to avoid cases when new content isn't available in the cache.  It maintains the use of caching for:

1. directory navigation (e.g. loading a file won't cause one API call per directory depth if the paths had already been loaded before).
2. directory content listing (e.g. running `ls` won't cause one API call per record if the records already existed).

Related to #102


Fenn-CS (approved):
If we can have an option to make this configurable in future it would be great. I mean it makes sense to crawl up with a cache but hit the API when crawling because a new file might exists. 

But it could also make sense to always use the cache if we wanted to.

This looks good.

Ty @slifty

Issue #25, opened 2022-06-16
Implement `REMOVE` and `RMDIR`


slifty:
This will allow SFTP clients to remove records and directories from their permanent archives.

The implementation must honor the (still to be determined / designed) setting on Permanent which specifies on a per-archive basis whether SFTP deletion is allowed.

slifty:
@meisekimiu In the rclone meeting today we identified this as a good issue for us to pair on so you'd have a chance to dig into the code base and make a change.

PR #110, opened 2023-03-22
Return failure status on remove operations

slifty:
This PR is a temporary patch to prevent hangs in the event that rclone has a reason to attempt to request the removal of a file or folder.  The remove will fail, but before we weren't even sending a response which would result in hangs from the sftp client's perspective.

Related to #25


Fenn-CS (approved):
Looks good! Thanks

Issue #111, opened 2023-03-22
Audit and fix temporary file calls


slifty:
We use a temporary file to store data being uploaded to permanent (since we need to know the file size before we can register a permanent record).  The flow that I had intended was:

1. Create (and open) the file when the SFTP `OPEN` request is made.
2. Write to the file for each SFTP `WRITE` request.
3. Close the file when the SFTP `CLOSE` request is made.

Unfortunately, as I learned while debugging #99, something about the above has been incorrectly implemented.  It could be that the library for creating the temporary files closes the file for us, or it could be some other issue.  Either way it is clear that I need to be very very careful with file operations and so I need to take a close look at the lifecycle of these temporary files.

slifty:
It looks like fs.close is [called by the `tmp` library](https://github.com/raszi/node-tmp/blob/da3bc1ac9a7880def4a3129c4f8bfe96a63bf99c/lib/tmp.js#L250) as part of the `removeCallback`, and so it may be as simple as just not calling close myself.

slifty:
I've opened an issue related to the leak #144 so I'm going to close out this one (which I consider to be more generic)

Issue #99, opened 2023-03-02
Multiple uploads caused some issues


slifty:
The root cause is not clear, but during a demo this afternoon an attempt to upload 6 `jpeg` files resulted in some errors.

This could be related to caching, it could be related to download name changes, or something else.

This issue is just a placeholder for me to dig in and learn more!

slifty:
I looked at the server logs and it isn't immediately clear what the problem is (the prod server has a log level of info so not all details are there).

That said, I'm seeing deletion requests which indicates to me that rclone thinks the files failed to upload properly.  I suspect this is related to #97 and might suggest that we need to do even more / need to somehow make sure that at *least* the size of a file is populated properly on permanent before we consider a file uploaded.

Other things to consider here is the what kinds of issues might be caused here by caching.

slifty:
I've spent some time debugging this situation.

Some notes:

1. It happens unpredictably -- and the more files being uploaded, the more likely it will happen.
2. When a file hangs it seems to indefinitely (I waited around 10 minutes before giving up).
3. The hang consistently occurs at the point of a `node-fetch` call, however the specific endpoint being accessed / the specific call does not seem to matter.  This is to say: the hang might happen when looking up a folder, looking up an archive record, creating an S3 URL.  The only consistent behavior I've seen is that the calls are to permanent resources (as opposed to the POST to amazon S3).  That being said this could be due to the fact that calls to permanent make up the vast majority (e.g. 99%) of the fetch calls associated with SFTP.

All this is to say: The root cause from the SFTP service's perspective is that certain API calls to the permanent backend are not resolving -- that is, they are neither succeeding or failing, but simply sitting open without timing out.

The unpredictable nature of the hangs indicates to me that the issue is not with any particular call, but rather something lower level (e.g. HTTP or database connection pools, locked resources, etc.)

I currently have two "obvious" paths of exploration:

1. *Looking into the HTTP library that we are using* (`node-fetch`).  This library could possibly have a bug in it that results in connections to silently fail in certain circumstances.  This is not compelling because (A) the library itself is already a major version ahead of what we are using, though we cannot upgrade due to node version, and (B) we are planning to get rid of the library anyway. I would prefer to use this as an opportunity to update the version of NodeJS to 18.x, as Node recently added native support for fetch and we could then remove the dependency.

2. *Looking into the permanent backend more closely*.  I should note that there are no obvious backend errors in the existing sentry or log files.

[...]

PR #112, opened 2023-03-22
Prevent double closing

slifty:
This PR removes a nasty bug that was coming from double-closing of the same file ID (the `tmp` library closes files as part of cleanup, so we don't need to close it directly).

Issue #111
Issue #99


cecilia-donnelly (approved):
As usual in this repo I'm not set up to test, but the explanation along the way and in the tickets makes sense, so +1 from me.

PR #114, opened 2023-03-22
Remove blocking filesystem calls

### LINKED ISSUE #113: Remove blocking filesystem calls
When debugging #99 I spent a few cycles looking into whether or not we were accidentally starving node's event loop.  The answer was no, but it is clear that we should never be using `sync` versions of fs calls in production.  There are a few instances of sync calls and they're easy enough to replace with async versions.

slifty:
This PR replaces the `sync` version of filesystem calls with their async alternatives.

Resolves #113

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -338,35 +338,52 @@ export class SftpSessionHandler {
     );
     const temporaryFile = this.openTemporaryFiles.get(handle.toString());
     if (temporaryFile) {
-      const { size } = fs.statSync(temporaryFile.name);
-      this.getCurrentPermanentFileSystem().createFile(
-        temporaryFile.virtualPath,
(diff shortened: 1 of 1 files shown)


src/classes/SftpSessionHandler.ts
```diff
@@ -741,19 +756,36 @@ export class SftpSessionHandler {
           case 'xa+': // append and read (file must not exist)
           case 'a': // append
           {
-            const temporaryFile = tmp.fileSync();
-            this.openTemporaryFiles.set(handle, {
-              ...temporaryFile,
-              path: filePath,
+            tmp.file((err, name, fd, removeCallback) => {
```
slifty, line 759: Just a note that these changes LOOK large but in reality I just moved all the previous logic into the callback.

Fenn-CS (approved):
Hurray on this!

PR #115, opened 2023-03-22
Ensure populated files

### LINKED ISSUE #105: Consider ensuring files have been processed before marking them as closed
When an ArchiveRecord is created on Permanent there are a bunch of processing steps that take place.  Some of these steps aren't really relevant to the SFTP service (e.g. thumbnail generation) but there are some that are critical, such as the generation of download URLs and the decoration of the file with an appropriate size.

Until an ArchiveRecord's original file has a downloadName and size, the file doesn't quite exist from the perspective of SFTP service since its stat data will be inaccurate, and the ability to open the file for read cannot happen.

One solution would be to modify the backend to populate these fields as part of the archive record creation, and while this would be ideal from SFTP's perspective it might not be realistic to implement in the near future.

As a stopgap measure the SFTP service can simply poll the backend until the file is correctly generated.  The nice thing about this approach is that if the backend does eventually change to immediately populate the file there won't *need* to be any change to the sftp service (it's just the polling logic won't get invoked).

slifty:
This PR adds logic to ensure that files are completely processed before allowing access to them.

The initial issue (105) is focused on waiting before marking as closed, but in reality any file could be accessed outside the context of a `close` event, and so the wait really needs to happen on `read` (be that a read based OPEN or an attempt to load file stats).

This PR also slips in some better logging.

Resolves #105

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -208,11 +208,41 @@ export class PermanentFileSystem {
     if (!isItemPath(requestedPath)) {
       throw new Error('Invalid file path');
     }
+    await this.waitForPopulatedOriginalFile(requestedPath);
     const archiveRecord = await this.loadArchiveRecord(
       requestedPath,
       overrideCache,
     );
     return getOriginalFileForArchiveRecord(archiveRecord);
+  }
+
+  public async waitForPopulatedOriginalFile(
+    requestedPath: string,
+    attemptNumber = 0,
+  ): Promise<void> {
+    if (attemptNumber >= 9) {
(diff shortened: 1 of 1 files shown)


Fenn-CS (approved):
LGTM

slifty:
I hear you!  10 minutes is a while, but I also don't have a sense yet of how long a file *COULD* process. If it finishes in 2 seconds then this would resolve in 2 seconds / If a file processes sooner then no problem.

That said, for a larger file (e.g. a 1GB video file) I could imagine processing taking longer.  "Giving Up" here would means the large file would be deleted and re-uploaded (once we support deletion), so I think it's better to err on the longer side.

PR #117, opened 2023-03-22
Bypass cache on miss

### LINKED ISSUE #116: Always re-load cache on a miss
In the continuous effort to improve caching (I knew #102 wouldn't be the last!) I'd like to update the cache logic to always re-load the parent directory if a given item is being searched for and results in a "miss".  This means the cache will be utilized when the file exists in the cache, but the cache won't be trusted / used if the file doesn't exist in the cache.

slifty:
This PR adds some logic to double-check if a file or folder is considered "missing" the first time we look for it.  With this addition we actually are able to simplify our cache management a bit (we no longer need to explicitly wipe out the cache data when creating new items, for instance).

Resolves #116

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -83,7 +83,10 @@ export class PermanentFileSystem {
     ];
   }
 
-  public async getItemType(itemPath: string): Promise<number> {
+  public async getItemType(
+    itemPath: string,
+    overrideParentCache = false,
+  ): Promise<number> {
     if (isRootPath(itemPath)
      || isArchiveCataloguePath(itemPath)
      || isArchivePath(itemPath)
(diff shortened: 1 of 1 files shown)


src/classes/PermanentFileSystem.ts
```diff
@@ -381,6 +388,29 @@ export class PermanentFileSystem {
...
+      overrideParentCache,
+    );
+    const targetArchiveRecord = parentFolder.archiveRecords.find(
+      (folder) => folder.fileSystemCompatibleName === archiveRecordName,
+    );
+
+    if (overrideParentCache && !targetArchiveRecord) {
+      throw new Error('The specified archive record does not exist');
+    }
+    return targetArchiveRecord ?? this.findArchiveRecordInParentDirectory(
```
Fenn-CS, line 407: @slifty some notes on why this is recursive would help, thanks!

slifty: Good call -- recursion deserves an explanation!  It's also probably the case that I should avoid recursion at all costs... but since I'm doing it i'll at least add the commment :D

slifty: I also was being too clever by half in this -- I've removed the `??` and made the logic flow more explicit / legible.

Fenn-CS (approved):
Looks good!

PR #120, opened 2023-04-12
Implement support for sftp RMDIR

slifty:
This PR adds support for the RMDIR handler -- meaning a user could delete directories from permanent via SFTP.

NOTE: Until the permanent SDK actually populates this setting from real API responses the setting will always be "false" / this code won't actually run.

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -8,6 +8,7 @@ import {
   getArchiveFolders,
   getFolder,
   getArchiveRecord,
+  getAuthenticatedAccount,
 } from '@permanentorg/sdk';
 import {
   deduplicateFileEntries,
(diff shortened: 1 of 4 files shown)


meisekimiu (approved):

src/classes/PermanentFileSystem.ts
```diff
@@ -188,6 +190,32 @@ export class PermanentFileSystem {
...
+    );
+    if (!account.isSftpDeletionEnabled) {
+      throw new Error('You must enable SFTP deletion directly in your account settings.');
+    }
+
+    if (isRootPath(requestedPath)) {
+      throw new Error('You cannot delete the root level folder.');
+    }
+    if (isArchiveCataloguePath(requestedPath)) {
+      throw new Error('You cannot the archive catalogue.');
```
Fenn-CS, line 205: Looks there's an omission here.

slifty: Whoops thank you!

PR #140, opened 2023-04-17
Throw more specific errors

### LINKED ISSUE #125: Throw more meaningful error messages
Right now our file system throws "Error"s for a wide variety of reasons (e.g. opening a file that doesn't exist, general API errors, trying to open a path as a folder when that path is a fike, etc).  It will be helpful to have more specific errors thrown (e.g. something like `FileDoesNotExistError`).

This should provide more context for things like #52, which will need to know why something failed in order to provide meaningful messages.

slifty:
This PR adds more specific error types to the PermanentFileSystem.

Please be sure to flag any error names that aren't intuitive to you or that you think would be better as something else.

Resolves #125

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,7 +5,8 @@
     },
     "plugins": [
       "@typescript-eslint",
-      "eslint-plugin-import"
+      "eslint-plugin-import",
+      "sort-exports"
     ],
     "extends": [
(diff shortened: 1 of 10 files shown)


Fenn-CS (approved):
Looks great! 

Thanks @slifty

PR #141, opened 2023-04-17
Fix parameter types for stat handlers

slifty:
This PR fixes an issue where the FSTAT handler was incorrectly attempting to treat a `handle` as a file path.  It also fixes an issue where other STAT handlers were incorrectly typing their path parameters as buffers when the ssh2 library specifies that it is passing strings.


cecilia-donnelly (approved):
I see what you're doing here but haven't run it myself.

PR #142, opened 2023-04-17
Ensure sftp calls always resolve

### LINKED ISSUE #121: All sftp events must resolve!
There are at least two handlers that log an error, but don't return a status.  This will result in hanging behavior for clients, as they expect some kind of response.

Two known instances are:

1. SSH_FXP_RENAME
2. SSH_FXP_SYMLINK

There may be more...

slifty:
This PR fixes a few SFTP handlers that were not providing a response to the SFTP client, which could result in client hangs in some circumstances.

Resolves #121


slifty:
@Fenn-CS ty!  Also no worries at all, this stuff was added a long long long time ago (maybe even a year ago) and we didn't quite know the ramifications of not sending a response at that time.

PR #147, opened 2023-04-19
Update github actions

slifty:
This PR updates our actions to latest (and configures dependabot to keep it that way going forward)


Fenn-CS (approved):
Thanks @slifty, looks okay

PR #154, opened 2023-04-20
Override node-fetch dependabot settings

slifty:
This PR does not allow dependabot to try to upgrade our `node-fetch` version to 3.x.


Fenn-CS (approved):
Great!

PR #177, opened 2023-05-19
Use s3 for temporary storage

### LINKED ISSUE #145: A user could create a DOS by uploading files that are too large
(This is related to #144)

We currently store in-transit data locally AND have no limits on how large a given uploaded file can be.  This means anybody could deny service by simply uploading a file that is larger than the locally configured storage space.

I believe the only solution here is to stop using local storage for temporary space; this would mean using S3 as the temporary storage location.  We might, separately, also want to create limits to how much data can be uploaded by a given user (this could be rooted in knowing how much storage space they have available on their account)

slifty:
This PR replaces the use of local storage for temporary upload progress with S3.

By making this change we prevent the risk of local file storage being filled.

Eventually the SDK and permanent backend will support multi-part upload, at which point we will be able to remove the S3 dependencies / direct integration with S3.

Resolves #145

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -3,13 +3,13 @@ import path from 'path';
 import fetch from 'node-fetch';
 import { v4 as uuidv4 } from 'uuid';
 import ssh2 from 'ssh2';
-import tmp from 'tmp';
 import { logger } from '../logger';
 import { generateFileEntry } from '../utils';
+import { NotFoundError } from '../errors';
+import { TemporaryFile } from './TemporaryFile';
 import type { AuthenticationSession } from './AuthenticationSession';
 import type { PermanentFileSystem } from './PermanentFileSystem';
 import type { PermanentFileSystemManager } from './PermanentFileSystemManager';
-import type { FileResult } from 'tmp';
 import type {
   Attributes,
   FileEntry,
@@ -22,10 +22,6 @@ import type {
 const SFTP_STATUS_CODE = ssh2.utils.sftp.STATUS_CODE;
 
 const generateHandle = (): string => uuidv4();
-
-interface TemporaryFile extends FileResult {
-  virtualPath: string;
-}
 
 export class SftpSessionHandler {
   private readonly sftpConnection: SFTPWrapper;
@@ -240,7 +236,7 @@ export class SftpSessionHandler {
     );
     const temporaryFile = this.openTemporaryFiles.get(handle.toString());
     if (!temporaryFile) {
-      logger.debug('There is no open temporary file associated with this handle', { reqId, handle });
+      logger.debug('There is no open file associated with this handle', { reqId, handle });
       logger.verbose(
         'Response: Status (FAILURE)',
         {
@@ -251,36 +247,31 @@ export class SftpSessionHandler {
       this.sftpConnection.status(reqId, SFTP_STATUS_CODE.FAILURE);
       return;
     }
-    fs.write(
-      temporaryFile.fd,
-      data,
-      0,
-      (err, written, buffer) => {
-        if (err) {
-          logger.verbose(
-            'Response: Status (FAILURE)',
-            {
-              reqId,
-              code: SFTP_STATUS_CODE.FAILURE,
-              path: temporaryFile.virtualPath,
-            },
-          );
-          this.sftpConnection.status(reqId, SFTP_STATUS_CODE.FAILURE);
-          return;
-        }
-        logger.debug('Successful Write.', { reqId, handle, written });
-        logger.silly('Written Data:', { buffer });
+    temporaryFile.append(data)
+      .then(() => {
+        logger.debug('Successful Write.', { reqId, handle });
+        logger.silly('Written Data:', { data });
         logger.verbose(
           'Response: Status (OK)',
           {
             reqId,
             code: SFTP_STATUS_CODE.OK,
-            path: temporaryFile.virtualPath,
+            path: temporaryFile.permanentFileSystemPath,
           },
         );
         this.sftpConnection.status(reqId, SFTP_STATUS_CODE.OK);
-      },
-    );
+      })
+      .catch(() => {
+        logger.verbose(
+          'Response: Status (FAILURE)',
+          {
+            reqId,
+            code: SFTP_STATUS_CODE.FAILURE,
+            path: temporaryFile.permanentFileSystemPath,
+          },
+        );
+        this.sftpConnection.status(reqId, SFTP_STATUS_CODE.FAILURE);
+      });
   }
 
   /**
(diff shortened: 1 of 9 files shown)


src/classes/SftpSessionHandler.ts
```diff
@@ -893,40 +866,38 @@ export class SftpSessionHandler {
           case 'xa+': // append and read (file must not exist)
           case 'a': // append
           {
-            tmp.file((err, name, fd, removeCallback) => {
-              if (err) {
+            const writeFile = new TemporaryFile(filePath);
```
slifty, line 869: This could throw an error (if S3 settings have not been provided) -- it would be good to catch that error here.

src/classes/TemporaryFile.ts
```diff
@@ -0,0 +1,114 @@
...
+      },
+    });
+    this.bucket = process.env.TEMPORARY_FILE_S3_BUCKET;
+    this.key = `${path}/${uuid()}`;
+    this.permanentFileSystemPath = permanentFileSystemPath;
+  }
+
+  public get url(): string {
+    if (this.s3Url === undefined) {
+      throw new Error();
```
Fenn-CS, line 60: What is this error?

cecilia-donnelly:
Tagged @liam-lloyd since he has been looking at multipart upload too.

src/classes/TemporaryFile.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+    if (process.env.AWS_ACCESS_KEY_ID === undefined) {
+      throw new SystemConfigurationError('AWS_ACCESS_KEY_ID must be populated in order to upload to s3.');
+    }
+    if (process.env.AWS_ACCESS_SECRET === undefined) {
+      throw new SystemConfigurationError('AWS_ACCESS_SECRET must be populated in order to upload to s3.');
+    }
+    if (process.env.TEMPORARY_FILE_S3_BUCKET_REGION === undefined) {
+      throw new SystemConfigurationError('TEMPORARY_FILE_S3_BUCKET_REGION must be populated in order to upload to s3.');
+    }
```
liam-lloyd, line 44: If this is how all files are going to be uploaded, would it make sense to check for these env vars at startup rather than in this constructor? If a missing env var means the service won't work at all I'd prefer to learn about that as soon as possible.

src/classes/TemporaryFile.ts
```diff
@@ -0,0 +1,114 @@
...
+      throw new Error('You cannot post data before opening the file');
+    }
+    const uploadPartCommand = new UploadPartCommand({
+      Bucket: this.bucket,
+      Key: this.key,
+      Body: fileData,
+      UploadId: this.uploadId,
+      PartNumber: this.partNumber,
+    });
+    const completedPart = await this.s3.send(uploadPartCommand);
```
liam-lloyd, line 90: I'm not sure where the right place is to check for this in this use case, but S3 won't accept a part smaller than 5 MB unless it's the last part.

src/classes/TemporaryFile.ts
slifty: Great idea! That said, I would also need / want to also check in the constructor for defensiveness and type safety purposes (this class can't assume that things were checked in an unrelated scope)

src/classes/TemporaryFile.ts
slifty: Oh dear... that's a big one!  I'm certain SFTP will be getting chunks smaller than 5 MB.  This would mean I need to write to temporary storage as well.

Thanks for the heads up; I'll need to do some refactoring.

slifty: For context on this limit (posting for me / historians): https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html

src/classes/TemporaryFile.ts
slifty: I realized this should not error but should just return an empty string if no URL is set.

liam-lloyd (approved):

src/index.ts
```diff
@@ -12,6 +13,19 @@ if ('SENTRY_DSN' in process.env
...
+if (process.env.AWS_ACCESS_KEY_ID === undefined) {
+  throw new SystemConfigurationError('AWS_ACCESS_KEY_ID must be populated in order to upload to s3.');
+}
+if (process.env.AWS_ACCESS_SECRET === undefined) {
+  throw new SystemConfigurationError('AWS_ACCESS_SECRET must be populated in order to upload to s3.');
+}
+if (process.env.TEMPORARY_FILE_S3_BUCKET_REGION === undefined) {
+  throw new SystemConfigurationError('TEMPORARY_FILE_S3_BUCKET_REGION must be populated in order to upload to s3.');
+}
+
```
Fenn-CS, line 28: This is exactly the same check in the `TemporaryFile` class. Can you wrap this check in a function and call it in these two places?

Not a hill I would die on though.

Fenn-CS (approved):

slifty:
I was about to write in the issue but realized the PR is probably the best place for it at this point -- In testing this out locally I realized that the "local buffer" implementation doesn't quite work out with the way sftp sends data (and, in fact, wouldn't have worked even without the local buffer, and maybe wouldn't even have worked necessarily in the original implementation!).

The SFTP protocol includes an `offset`:

```
`offset' is the offset (in bytes) from the
   beginning of the file where to start writing, and `data' is the data
   to be written.

   The write will extend the file if writing beyond the end of the file.
   It is legal to write way beyond the end of the file; the semantics
   are to write zeroes from the end of the file to the specified offset
   and then the data.  On most operating systems, such writes do not
   allocate disk space but instead leave "holes" in the file.   ```
   
Which is to say, writes are done in parallel AND are not guaranteed to be sequential.  I'm currently working to figure out a good way to handle this.

slifty:
It looks like the SFTP protocol (v3) does not provide a way to signal to clients that writes must be sequential.

I think this means that, well, we need to support non-sequential writes and know how to honor the `offset`.

The trouble here is we now have some conflicting constratings:

1. We want a finite cap on the amount of local data that can be stored per connection.
2. S3 requires us to write 5MB chunks.

If data is being written sequentially we have no issue: just append to a single local temporary file and when it gets large enough, upload it.

Since data is NOT being written sequentially, we have no way of ensuring that we will get a 5MB chunk.

For example, one could imagine imagine an SFTP client that breaks a 100MB file into twenty separate 5MB segments, and sends parts of each segment in a round robin style.  This would functionally result in storing all 100MB in local temporary storage before writing anything to S3!

It does appear that rclone is writing sequentially, by the way, it's just that we need to appreciate that this is not guaranteed.

I'm going to reflect on alternate approaches given these revelations.  One approach could be to use s3 for even the "temporary" storage as a series of individual, small files that are indexed by their offsets and ultimately get coalesced.  I am concerned that this will lead to massive overhead in creating so many tiny S3 objects (and I am also unsure if S3 prices per object, which I could see making this tactic very expensive).

Another approach could be to simply implement our current plan, knowing that generally clients are *generally* writing sequentially.  We could also add limits to the allowed local storage (but those limits might be higher than 5MB).  This could represent a "sweet spot" of preventing DDOS risk but also could lead to confusing file upload failures for users if their SFTP client has an obnoxious write fragmentation algorithim

slifty:
Marking this as draft for now since there have been enough odd curve balls that it warrants stepping back and making sure this is still the approach we want to take.

I also just spoke with @kfogel and I'm going to at least update the current (full tmp storage) implementation to properly account for offset.

PR #189, opened 2023-06-14
Correct local remote addresse in .env.example

Fenn-CS:
Copying and pasting the existing local remote from `.env.example` (`http://local.permanent.org/api`) could lead to some unwarranted debugging.


meisekimiu (approved):

Fenn-CS:
Thanks @meisekimiu

Issue #175, opened 2023-05-15
Refresh FusionAuth token 


Fenn-CS:
When the file uploads would exceed an hour the backend would fail with `Firebase\JWT\ExpiredException` on `GET /folder/getWithChildren`

Leading to 

```
2023/05/15 16:42:37 ERROR : nzmgdjdhzj: Failed to copy: Put mkParentDir failed: mkdir "/archives/rclone QA 1 (dev) (07av-0000)/My Files" failed: sftp: "" (SSH_FX_FAILURE)
2023/05/15 16:42:37 ERROR : nzndqkkgwb: Failed to copy: Put mkParentDir failed: mkdir "/archives/rclone QA 1 (dev) (07av-0000)/My Files" failed: sftp: "" (SSH_FX_FAILURE)
2023/05/15 16:42:37 ERROR : nynthjaqhy: Failed to copy: Put mkParentDir failed: mkdir "/archives/rclone QA 1 (dev) (07av-0000)/My Files" failed: sftp: "" (SSH_FX_FAILURE)
2023/05/15 16:42:38 ERROR : obmqbiapma: Failed to copy: Put mkParentDir failed: mkdir "/archives/rclone QA 1 (dev) (07av-0000)/My Files" failed: sftp: "" (SSH_FX_FAILURE)

```

On the sftp client.

Apparently the FusionAuth token expires after one hour and so it needs to get refreshed.

Sentry : https://permanentorg.sentry.io/issues/3899931004/?environment=dev&project=3663562&query=is%3Aunresolved&referrer=issue-stream&stream_index=0

kfogel:
See also [related Slack thread](https://app.slack.com/client/TBBFM3TEY/C01ARGVA849/thread/C01ARGVA849-1684170022.916489).

kfogel:
Just moving some thoughts/commentary from team meeting notes docs over to here:

One option would be to just make the default tokens last longer (but "ick", for all the usual reasons).

Another option would be for the SFTP Service to refresh the FusionAuth token.  In that case, refresh tokens would need to be enabled.  Would the refresh token then also need to last longer than 12 hours, to allow for really long uploads?  (Or can a refresh token be used to get a newer refresh token?  That would be kind of suprising if it could, but who knows.)

See https://learn.microsoft.com/en-us/linkedin/shared/authentication/programmatic-refresh-tokens - could make them last a long time!

[...]

PR #192, opened 2023-06-20
Refresh authToken when it expires

Fenn-CS:
Access tokens have lifetimes that might end before the user wants to end their authenticated session. Hence, we need to refresh the user access token after an expiry detection.

Resolves: https://github.com/PermanentOrg/sftp-service/issues/175


src/classes/AuthenticationSession.ts
```diff
@@ -178,4 +181,18 @@ export class AuthenticationSession {
       this.authContext.reject();
     });
   }
+
+  private obtainNewAuthTokenUsingRefreshToken(): void {
+    this.fusionAuthClient.exchangeRefreshTokenForAccessToken(this.refreshToken, '', '', '', '')
+      .then((clientResponse) => {
+        this.authToken = clientResponse.response.access_token ?? '';
+      })
```
jasonaowen, line 189: You probably need to also update the refresh token here - ideally, the response should include both a new access token and a new refresh token, so that the authentication server can defend against token replay attacks. I don't know if FusionAuth specifically does that, but if it gives you a refresh token in the response - even if in practice it's the same - you should update your copy.

src/classes/AuthenticationSession.ts
```diff
@@ -178,4 +181,18 @@ export class AuthenticationSession {
       this.authContext.reject();
     });
   }
+
+  private obtainNewAuthTokenUsingRefreshToken(): void {
```
jasonaowen, line 185: I know this isn't yet being called anywhere, but if I may suggest, you probably want a method like `getAccessToken` that examines the expiration date - which you'll also need to track - and will transparently refresh if the access token is too old.

Ideally, this would be provided by your client library, but unfortunately this is still an open issue upstream: https://github.com/FusionAuth/fusionauth-typescript-client/issues/74 .

I haven't looked closely enough to say if this is a worthwhile tradeoff, but as an alternative to implementing all of this security code yourself, you may want to consider switching to another client library. Unless you're doing anything FusionAuth-specific, you should hopefully be able to use any OAuth/OIDC client library.

If you do decide to switch clients, I'd suggest doing so as its own stand-alone PR, before implementing this additional functionality.

jasonaowen (commented):
This approach seems reasonable, although there may be an alternative - see below.

src/classes/AuthenticationSession.ts
Fenn-CS: Thanks for the input Jason.

Currently, I can't get Fusionauth to include refresh tokens in its responses. (Despite the checkbox being checked in the admin, not sure if you know anything about that too)

Fenn-CS:
So apparently, even if the `generateRefreshToken` is enabled on the `FusionAuth` dashboard, a request has to be made to have the `refreshToken` included in the initial login response.

Something like

```ts
{
this.fusionAuthClient.login({
      loginId: this.authContext.username,
      password,
      generateRefreshTokens: true,
    })
 }).then(() => {})
 ```

in 

https://github.com/PermanentOrg/sftp-service/blob/a9ca4574043b3a706e6474ebe59a39367b4a9a31/src/classes/AuthenticationSession.ts#L47-L50

However, the `LoginRequest` type (interface) does not have that, neither does its parent classes do.

https://github.com/FusionAuth/fusionauth-typescript-client/blob/6f8d2e59362578e2f76b2e520818e3bec99e2a56/src/FusionAuthClient.ts#L6217-L6222

Next step is to expirement with `OAuth/OIDC` library to see these limitations in FusionAuth's typescript client can be bypassed.

File: src/classes/AuthenticationSession.ts (hunk 1 above)
Comment (jasonaowen) [reply src/classes/AuthenticationSession.ts:189]: You need to include the `offline_access` scope in your initial request: https://fusionauth.io/docs/v1/tech/oauth/tokens#refresh-token . It's been a bit since I've seen their admin page, but I'd expect that checking that box *allows* the selected client to request refresh tokens, rather than configuring FusionAuth to give them out unasked-for.

**Comment by jasonaowen** [comment]
> https://github.com/PermanentOrg/sftp-service/blob/a9ca4574043b3a706e6474ebe59a39367b4a9a31/src/classes/AuthenticationSession.ts#L47-L50

It is concerning that this service is requesting passwords from users! I don't know how other rclone backends handle OAuth, but I worry that the inspiration for doing so here came from the back-end. That was a compatibility shim I added to the back-end until the front-end was able to delegate its sign-in screen to FusionAuth - it is not the recommended approach, as user passwords should be exposed to as few services as possible, and ideally none of them.

What I would have expected is for signing in to go through some kind of web browser flow, probably the [device authorization grant](https://oauth.net/2/grant-types/device-code/), so that the user can provide their password exclusively to FusionAuth, as well as go through whatever MFA steps they have configured. Taking the responsibility of doing password and MFA into this app makes the process less secure and requires more code.

> Unless you're doing anything FusionAuth-specific, you should hopefully be able to use any OAuth/OIDC client library.

Unfortunately, it looks like you *are* doing FusionAuth-specific things. So long as you use their vendor-specific APIs rather than the standard OAuth/OIDC APIs, you'll have to continue to use their client library. I'd suggest moving away from using their vendor-specific APIs.

**Comment by Fenn-CS** [comment]
> It is concerning that this service is requesting passwords from users! I don't know how other rclone backends handle OAuth, but I worry that the inspiration for doing so here came from the back-end. That was a compatibility shim I added to the back-end until the front-end was able to delegate its sign-in screen to FusionAuth - it is not the recommended approach, as user passwords should be exposed to as few services as possible, and ideally none of them.

I am sure you know this from looking at the code, but the passwords are still passed to the FusionAuth api directly not passed via the back-end. 

Now, I can't remember the specific challenges that Dan faced with trying to authenticate directly with FusionAuth without collecting passwords.

> Unfortunately, it looks like you are doing FusionAuth-specific things. So long as you use their vendor-specific APIs rather than the standard OAuth/OIDC APIs, you'll have to continue to use their client library. I'd suggest moving away from using their vendor-specific APIs.

On this, I [attempted avoid the FusionAuth typescript client](https://github.com/PermanentOrg/sftp-service/pull/194) in a bit to get this `login` method return with `refreshToken` but eventually realized that, the we have no control over that via the API calls at least according to FusionAuth docs.

https://fusionauth.io/docs/v1/tech/apis/login#request 

> The Login API is used authenticate a user in FusionAuth. The issuer of the One Time Password will dictate if a JWT or a Refresh Token may be issued in the API response.

So at this point, looks like I am stuck with the FusionAuth dashboard again or maybe one really needs to change the authentication mechanism to move away from this login type completely.

**Comment by jasonaowen** [comment]
> > It is concerning that this service is requesting passwords from users! [...]
> 
> I am sure you know this from looking at the code, but the passwords are still passed to the FusionAuth api directly not passed via the back-end.

Yes, but my point is that this service should not have access to user passwords, either - only FusionAuth should. By accepting passwords, this service becomes a potential vector for compromising a user's credentials. I'm not saying that's likely, but it's an avoidable risk.

> On this, I [attempted avoid the FusionAuth typescript client](https://github.com/PermanentOrg/sftp-service/pull/194) in a bit to get this `login` method return with `refreshToken` but eventually realized that, the we have no control over that via the API calls at least according to FusionAuth docs.
> 
> https://fusionauth.io/docs/v1/tech/apis/login#request
> 
> > The Login API is used authenticate a user in FusionAuth. The issuer of the One Time Password will dictate if a JWT or a Refresh Token may be issued in the API response.
> 
> So at this point, looks like I am stuck with the FusionAuth dashboard again or maybe one really needs to change the authentication mechanism to move away from this login type completely.

Using the FusionAuth APIs via direct HTTP calls would be the worst of both worlds, unfortunately. If you must use their proprietary APIs, you'll be best off with their client.

To address the specific concern you have, from reading the docs, it appears that in order to get a refresh token via the login API you must pass your application ID:

>  Because a refresh token is per user and per application, when [the `applicationId`] parameter is not provided a refresh token will not be returned.

https://fusionauth.io/docs/v1/tech/apis/login#request-body

**Comment by slifty** [comment]
@jasonaowen I can talk a bit about why we chose to implement login the way we did, though I think that will be best explained by voice at least initially (I do allude to some of the reasons here: https://github.com/PermanentOrg/sftp-service/issues/175#issuecomment-1636290459 -- ultimately it is because rclone requires pre-configured credentials to connect to a remote host.  It *can* be set up to use the `ssh-agent` but I'm not certain how ssh-agent handles oauth at the time of writing this).

Ultimately I don't think that it will be within scope to develop support for oauth as part of this ticket.

It does appear that `login()` doesn't support refresh tokens, which makes it seem to me that the only solution that doesn't involve changing the auth mechanism is to extend the life of the token.  I kicked off a conversation about this over in #175

**Comment by slifty** [comment]
Jason pinged me to point out that we missed a key line of his point!

> To address the specific concern you have, from reading the docs, it appears that in order to get a refresh token via the login API you must pass your application ID

@Fenn-CS it looks like we may well have a solution!  Could you take a look at that?

**Comment by Fenn-CS** [comment]
> Jason pinged me to point out that we missed a key line of his point!
> 
> To address the specific concern you have, from reading the docs, it appears that in order to get a refresh token via the login API you must pass your application ID
> 
> @Fenn-CS it looks like we may well have a solution! Could you take a look at that?

So I replaced the login method with a custom client call that accepts applicationId in https://github.com/PermanentOrg/sftp-service/pull/194

This is what the response looks like : 

```
{
  "token": "eyJhbxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
  "tokenExpirationInstant": 1692728941867,
  "user": {
    "active": true,
    "connectorId": "e3306678-a53a-0000-1111-1c96f36dda72",
    "data": {},
    "email": "nfebe@withheld.com",
    "fullName": "Nfebe",
    "id": "b3c6c1b6-ffff-cccc-9da2-91a16016e11a",
    "insertInstant": 1690112959586,
    "lastLoginInstant": 1690136941867,
    "lastUpdateInstant": 1690112959586,
    "memberships": [],
    "passwordChangeRequired": false,
    "passwordLastUpdateInstant": 1690000959595,
    "preferredLanguages": [],
    "registrations": [],
    "tenantId": "45586253-3e3b-xxx-xxx-44eae469ec35",
    "twoFactor": {
      "methods": [],
      "recoveryCodes": []
    },
    "usernameStatus": "ACTIVE",
    "verified": true
  }
}
```

Conclusion: Passing `applicationId` does not lead the `refreshToken` being returned.

cc: @slifty

File: src/classes/AuthenticationSession.ts [hunk 3]
```diff
@@ -15,10 +15,16 @@ enum FusionAuthStatusCode {
...
 
+  public refreshToken = '';
+
   public readonly authContext;
 
+  private authTokenExpiresAt = 0;
+
   private readonly fusionAuthClient;
 
+  private readonly fusionAuthAppId = process.env.FUSION_AUTH_APP_ID ?? '';
```
slifty, line 26: This should either be a static attribute of an appropriate class (this class is fine) *OR* it should be passed as part of the constructor (e.g. `sftpFusionAuthAppId`) <-- also I'd specify in the name of the variable AND the env variable that it is intended to be the **sftp** fusion auth app ID (since there is more than one application id a dev could pick from)

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+
+    if (!AuthenticationSession.sftpFusionAuthClientSecret) {
+      logger.error(
+        'Cannot obtain new access token without sftp client secret.',
+      );
+      return;
+    }
+
+    this.fusionAuthClient.exchangeRefreshTokenForAccessToken(
+      this.refreshToken,
```
Fenn-CS, line 61: Upon hitting the endpoint directly with `rested` (postman-like tool) I realize that these parameters are required anyways. At least, the `clientId` and `clientSecret` for this use-case.

So the "issue" with the  fusionauth sdk is both poor documentation and poor use of typing but yay it works~

Fenn-CS:
Notes for testing

- Token expiration time in FusionAuth for sftp-local is set to 6 minutes (and can be changed while testing to a desired value).
- Our code will refresh the token when it has less than 5 minutes remaining (So with the current setting the tester should expect a new token roughly every 1 minute and 1 second).
- During testing, monitor the logs to observe a new token being obtained approximately every minute.
- If the expiration time in FusionAuth is set to 1 hour, expect a new token every 45 minutes, so a new token is obtained every `time_set_in_fusion_auth - 5 minutes`

_The short 6-minute expiration is for local testing only._

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+      );
+      return;
+    }
+
+    if (!AuthenticationSession.sftpFusionAuthClientSecret) {
+      logger.error(
+        'Cannot obtain new access token without sftp client secret.',
+      );
+      return;
+    }
```
jasonaowen, line 58: I appreciate these checks, but I think they're occurring too late, and returning early from a method that returns nothing does not clearly communicate that the service is in an invalid state.

Instead, I'd like to confirm that these environment variables are present during server startup, and exit if they're missing. (Are there any other environment variables like that?) I've previously used the [require-env-variable package](https://www.npmjs.com/package/require-env-variable) to do so, and would suggest that here.

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+
+    this.fusionAuthClient.exchangeRefreshTokenForAccessToken(
+      this.refreshToken,
+      AuthenticationSession.sftpFusionAuthClientId,
+      AuthenticationSession.sftpFusionAuthClientSecret,
+      '',
+      '',
+    )
+      .then((clientResponse) => {
+        this.authToken = clientResponse.response.access_token ?? '';
```
jasonaowen, line 68: This `?? ''` is not safe.

The [SDK is unhelpful here](https://github.com/FusionAuth/fusionauth-typescript-client/blob/a6b1fd4844fd7242c06de812353b493c1e27979e/src/FusionAuthClient.ts#L8802-L8811), as it marks every single attribute as optional, which means that we need to ensure that the response has all the attributes we need. If the API actually did return with a "successful" response that did not include an access token, we need to consider it an error here, not later when we try to use it.

Likewise for all the other nullish coalescing operators throughout this PR.

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+    }
+
+    this.fusionAuthClient.exchangeRefreshTokenForAccessToken(
+      this.refreshToken,
+      AuthenticationSession.sftpFusionAuthClientId,
+      AuthenticationSession.sftpFusionAuthClientSecret,
+      '',
+      '',
+    )
+      .then((clientResponse) => {
```
jasonaowen, line 67: I believe this would be more readable with `async`/`await` than using the Promise API. You're using `async`/`await` elsewhere in this PR; is there a reason not to here?

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+        const message = isPartialClientResponse(clientResponse)
+          ? clientResponse.exception.error_description
+          : '';
+        logger.warn(`Error obtaining refresh token : ${message}`);
+        this.authContext.reject();
+      });
+  }
+
+  public tokenExpired(): boolean {
+    const expirationDate = new Date(this.authTokenExpiresAt);
```
jasonaowen, line 87: Let's store the expiration time as a `Date` object when we first calculate it, instead of making a new one every time we check for expiration.

src/classes/SftpSessionHandler.ts
```diff
@@ -1070,6 +1070,12 @@ export class SftpSessionHandler {
   }
 
   private getCurrentPermanentFileSystem(): PermanentFileSystem {
+    if (
+      this.authenticationSession.tokenExpired()
+      || this.authenticationSession.tokenWouldExpireSoon()
```
jasonaowen, line 1075: This is not where I'd expect this check to live, as it is not obvious to me that it will be made before each and every API request. I'm not sure where it would better be placed; perhaps in `getClientConfiguration`? It would be nice if the Permanent SDK helped us out here. @slifty what do you think?

jasonaowen (changes requested):
Thanks for working on this, @Fenn-CS! Very cool to see so much progress.

src/classes/AuthenticationSession.ts
slifty: +1 to checking immediately during server startup.

As part of that, it might be cleaner to have these values passed as part of a configuration to a constructor (and then have type protection for ensuring they exist rather than having to check in runtime).

src/classes/AuthenticationSession.ts
slifty: Not just an issue of readability -- the core functionality of the function is to generate and set an instance attribute.  By hiding the fact that it's async in the function signature we're creating room for race conditions.

This function *must* be marked as async!

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+    return expirationDate <= new Date();
+  }
+
+  public tokenWouldExpireSoon(minutes = 5): boolean {
```
slifty, line 91: Some thoughts here:

(1) It seems like we could combine these two flavors of "stale" into a single function.  If a token has only 4 minutes left or a token expired (has negative minutes left) in either case it would be less than the `minutes` and as far as I can tell the difference between "stale" vs "expired" has no logical impact at this point.

(2) `minutes` is generic (which makes the check at the end a bit hard to parse) -- maybe something like `staleThreshold`

(3) It is odd to me for duration to be passed in terms of minutes.  I'd make this seconds or milliseconds, which is more intuitive in the land of JS.

src/classes/AuthenticationSession.ts
```diff
@@ -45,41 +108,59 @@ export class AuthenticationSession {
   }
 
   private processPasswordResponse([password]: string[]): void {
+    if (!AuthenticationSession.sftpFusionAuthAppId) {
```
slifty, line 111: As discussed -- since this is required for the class to operate let's just make it part of the constructor and we can elevate these checks to `server.ts` or `index.ts`.

src/classes/AuthenticationSession.ts
```diff
@@ -45,41 +108,59 @@ export class AuthenticationSession {
...
           return;
-        case FusionAuthStatusCode.SuccessNeedsTwoFactorAuth:
+        }
+        case FusionAuthStatusCode.SuccessButUnregisteredInApp: {
+          const userId: string = clientResponse.response.user?.id ?? '';
+          this.registerUserInApp(userId)
+            .then(() => { this.processPasswordResponse([password]); })
+            .catch((error) => {
+              logger.warn('Error during registration and authentication:', error);
+              this.authContext.reject();
```
slifty, line 140: I forget if `reject` takes an optional message but if so we should communicate the reason for rejection (this should be done in the other locations too FWIW, I missed these when doing #212).

src/classes/SftpSessionHandler.ts
slifty: Agreed @jasonaowen!  This check should be done *when accessing the authToken* -- I expand on this concept in more detail in another comment on this review.

src/fusionAuth.ts
```diff
@@ -8,6 +8,8 @@ export const getFusionAuthClient = (): FusionAuthClient => new FusionAuthClient(
 export interface PartialClientResponse {
   exception: {
     message: string;
+    error?: string;
```
slifty, line 11: Just a sanity check: is it possible for FusionAuth to have an exception without an error / error_description?

Would it make sense to have a separate type / interface of `PartialClientErrorResponse` that more explicitly defines what we're expecting?

src/classes/AuthenticationSession.ts
```diff
@@ -32,6 +42,59 @@ export class AuthenticationSession {
...
+        // The exchange refresh token for access token endpoint dose not return a timestamp,
+        // it returns expires_in in seconds.
+        // So we need to create the timestamp to be consistent with what is first
+        // returned upon initial authentication
+        this.authTokenExpiresAt = Date.now() + (clientResponse
+          .response.expires_in ?? 1 * 1000);
+        logger.info('New access token obtained');
+      })
+      .catch((clientResponse: unknown) => {
+        const message = isPartialClientResponse(clientResponse)
```
slifty, line 78: Could we update this to have more explicit blocks that handle types of errors?

`clientResponse` is just one of several reasons a promise could reject, and so the `catch` should be `catch((error: unknown)) =>` and then we should have `if(isPartialClientResponse) { ... }` and an explicit case for other types of error.

slifty (changes requested):
Thanks for the progress on this!  I identified some changes I think would make this implementation even better / safer!

src/classes/AuthenticationSession.ts
slifty: I don't know if my comment got lost in GitHub UI or I never made it but I just wanna underscore that as written this introduces a race condition and we *must* use async / await since the calling code relies on the result of this async call.

src/classes/AuthenticationSession.ts
```diff
@@ -13,10 +13,20 @@ enum FusionAuthStatusCode {
 }
 
 export class AuthenticationSession {
+  private static readonly sftpFusionAuthAppId = process.env.FUSION_AUTH_SFTP_APP_ID ?? '';
```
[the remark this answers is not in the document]
Fenn-CS: We won't need them as instance properties with https://www.npmjs.com/package/require-env-variable

src/fusionAuth.ts
Fenn-CS: Yes, sometimes they return just the message. I checked their response interface and `error` is optional, I wonder why they would do that.

Fenn-CS: Looking at your second comment, this is most likely because I was catching the `clientResponse` and not the exception, but I wonder if it makes any difference because at the end `clientResponse` is just a function-level scoped var name.

slifty: Right.  Regarding that second point: you're catching any possible exception -- client response is just one type of exception, which is why `clientResponse` is an inaccurate name for that variable name since it could be anything (hence being unknown!)

So I'd rename it to `err` and then have type-check based handling for anticipated error types (e.g. the client response), and wrap up with a generic "unknown error" response.

src/classes/AuthenticationSession.ts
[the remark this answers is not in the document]
slifty: @Fenn-CS I hear you, but I think it would be more explicit / better practice to have them as instance properties and require them as part of the construction of an `AuthenticationSession`.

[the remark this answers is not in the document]
Fenn-CS: @slifty all we need to with the `require-env-variable` package is to import that variables and it handles the rest including failure and exit in case of env var absence.

I mean

```js
const {
  FUSION_AUTH_SFTP_APP_ID,
  FUSION_AUTH_SFTP_CLIENT_ID,
  FUSION_AUTH_SFTP_CLIENT_SECRET,
} = requireEnv(
  'FUSION_AUTH_SFTP_APP_ID',
  'FUSION_AUTH_SFTP_CLIENT_ID',
  'FUSION_AUTH_SFTP_CLIENT_SECRET',
);
```

Since we are not using any of these `env` vars upper in the stack, those would become unused imports, unless we want to force use them without need to annotate the import so that it is ignored by the linter. 

I think it's still okay for the system to exit at the point where it needs something and it does not exist it must not be at the very top.

src/classes/AuthenticationSession.ts
Fenn-CS: I can add this in a separate commit once it's clear to me that it takes on optional message. 

For now, the type hints tell me it takes arrays of "authentication types".

https://github.com/PermanentOrg/sftp-service/blob/main/src/classes/SshConnectionHandler.ts#L37

src/classes/AuthenticationSession.ts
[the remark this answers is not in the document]
slifty: I hear you from an implementation perspective, but I want to be sure what I'm saying is clear because I'm talking about code design!

This class requires these values to function, and having these values as part of the construction signature is *more explicit* and signals that requirement to any developer that might use the class.  This does a few things:

1. If there was ever a need to use this class to authenticate with multiple fusion app IDs, we could have different instances to handle those different IDs.

2. It decouples the validation of these values from this class (to your point, validation right now is handled by the requireEnv library, but imagine a situation where validation was more involved)

3. It allows us to recognize a failure in configuration instantly on deployment -- which is very important in a production environment.

What I'm proposing is that you require these things at `server.ts` and then pass them from class to class -- the Ssh handler class would need to know what the App ID, Client ID, and Client Secret for its SFTP subcomponent are supposed to be.

As of now there is no way, for example, (without inspecting or searching the entire code base) for a developer to realize that these environment variables are required in order for the SshSessionHandler to function!  That's not ideal and contributes to tech debt over time.

NOTE: if you and / or @jasonaowen think that explicitness isn't worth the trouble in this case then I totally understand!  I just wanted to make sure I was clear in explaining that what I was suggesting wasn't just to address implementation challenges.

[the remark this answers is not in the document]
Fenn-CS: Thanks @slifty that's much clearer and can be done without much hassle. I would add this final bit of context to know your thoughts, then go ahead. So, what `requireEnv` actually does is that, you won't be able to build the source or `run dev` at all if the environment variables are not present. Given that, and that, the class that directly depends on these `AuthenticationSession` now requires these as instance values passed in the constructor, do you still think it's important for us to pass these values about 2 - 3 classes down just to get in to the `AuthenticationSession` class?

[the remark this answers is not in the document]
Fenn-CS: You know what? It's one class down, I would just do that!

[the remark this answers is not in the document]
slifty: That's good context (that run dev would immediately break) -- and that does address the biggest concern from a practical perspective.

The code design question is a little bit more of a philosophical question and I am completely open to the idea that what I'm suggesting is not worth the tradeoff in terms of code complexity.

All this is to say: at this point it sounds like you totally grok what I was trying to explain, and I defer to you / the group on what you think is best!  Thank you for coming to my TED talk 😆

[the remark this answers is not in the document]
Fenn-CS: It's done @slifty so please kindly review, I won't mark this as resolved because we had this conversation I would leave that to you.

src/classes/AuthenticationSession.ts
slifty: It looks like it does *NOT* take an additional message!  I'm marking this as resolved.

src/classes/AuthenticationSession.ts
```diff
@@ -13,25 +13,94 @@ enum FusionAuthStatusCode {
...
+        // it returns expires_in in seconds.
+        // So we need to create the timestamp to be consistent with what is first
+        // returned upon initial authentication
+        this.authTokenExpiresAt = new Date(
+          Date.now() + (clientResponse.response.expires_in ?? 1 * 1000),
+        );
+        logger.info('New access token obtained :', clientResponse.response);
+      } else {
+        logger.warn('No refresh token in response :', clientResponse.response);
+        this.authContext.reject();
```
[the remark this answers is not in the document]
slifty: RELATED NOTE BUT PLEASE IGNORE IT: I'm now realizing that we've actually expanded the scope of `AuthenticationSession` to being BOTH the class that handles the establishment of an SFTP authentication session *and* being a class that manages FusionAuth tokens.  It is definitely conceptually confusing to have both in one place, and that leads to attempting to do things like this.

It might be best to have a new class called `AuthTokenManager`, which is solely responsible for managing the auth token / refresh token for the authentication session, contains the refresh token logic, and has no risk of accidentally using irrelevant (and dangerous) class attributes.

This would also be the only thing that has to be passed to the sftp session as well -- which would fix an original mistake of... why is an authSession passed to `SftpSessionHandler` in the first place!

That said, this feels out of scope.  I think what I'll do is actually just open an issue for this once the code is merged

slifty (changes requested):
Phewf this is so close thank you for all the updates.

Some of my comments are style / naming suggestions which please feel free to take or leave.  That said, there are some more important notes here:

1. A small bug around time calculation (I think)
2. We can't call `authContext.reject` when refreshing auth tokens

That second issue is a big one, and even led me to realize that there would be merit to extracting the auth token management from this class entirely (which I'm suggesting we do as a separate issue rather than addressing in this PR).

src/classes/AuthenticationSession.ts
```diff
@@ -13,25 +13,94 @@ enum FusionAuthStatusCode {
...
+        message = error instanceof Error ? error.message : JSON.stringify(error);
+      }
+      logger.warn(`Error obtaining refresh token: ${message}`);
+      this.authContext.reject();
+    }
+  }
+
+  private tokenWouldExpireSoon(seconds = 300): boolean {
+    const currentTime = new Date();
+    const timeDifferenceSeconds = (
```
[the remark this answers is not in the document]
Fenn-CS: This returns a boolean, `remainingTokenLife` sounds like it should return TTL or time in practice. What do you think?

src/classes/AuthenticationSession.ts
```diff
@@ -13,25 +13,94 @@ enum FusionAuthStatusCode {
...
+        // returned upon initial authentication
+        this.authTokenExpiresAt = new Date(
+          Date.now() + (clientResponse.response.expires_in ?? 1 * 1000),
+        );
+        logger.info('New access token obtained :', clientResponse.response);
+      } else {
+        logger.warn('No refresh token in response :', clientResponse.response);
+        this.authContext.reject();
+      }
+    } catch (error: unknown) {
```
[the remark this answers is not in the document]
Fenn-CS: Umm, the block is 28 lines. 7 lines are comments, 7 lines are the actual call, 12 lines are things that directly depends on the call succeeding and hence bundled in the try catch. 

Alternative is defined a null response out the try-catch and only catch the and then handle process checks after the catch which makes it less try-catch counter intuitive IMO.

Not marking as resolve, if there's more context to be thrown in. Thanks @slifty

src/classes/AuthenticationSession.ts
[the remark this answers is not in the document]
Fenn-CS: Did not rename, so not marking as resolved, so that discussion can continue if need be.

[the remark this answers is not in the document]
slifty: Sorry for being ambiguous!  I was talking about renaming the variable `timeDifferenceInSeconds` (not the function name!)

Basically that variable is storing the remaining token life -- which, yes, is calculated in terms of a difference in time, but the current variable name doesn't indicate what it actually *means*

src/classes/AuthenticationSession.ts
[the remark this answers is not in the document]
slifty: Just to be clear your try / catch block would still stop processing in the event of a failure -- so the code that relies on success does NOT need to be wrapped in the entire try block.  Right now the try block throws several errors which immediately get caught and re-thrown in the catch block.

Here's what I mean as an alternative (typed in github so may well have issues):

```
let clientResponse;
try {
      /**
       * Fusion auth sdk wrongly mandates last two params (scope, user_code)
       * hence the need to pass two empty strings here.
       * See: https://github.com/FusionAuth/fusionauth-typescript-client/issues/42
       */
      const clientResponse = await this.fusionAuthClient.exchangeRefreshTokenForAccessToken(
        this.refreshToken,
        this.fusionAuthClientId,
        this.fusionAuthClientSecret,
        '',
        '',
      );
} catch (error: unknown) {
      if (isPartialClientResponse(error)) {
        throw new AuthTokenRefreshError(`Error returned by FusionAuth while obtaining refresh token: ${message}`);
      }
      if (error instanceof Error) {
        throw new AuthTokenRefreshError(`Error generated when calling by FusionAuth to obtaining refresh token: ${error.message}`);
      }
      console.log(error);
      throw new AuthTokenRefreshError('Unexpected throw when calling FusionAuth to obtain refresh token');
}

if (clientResponse.response.access_token) {
  this.authToken = clientResponse.response.access_token;
  if (clientResponse.response.expires_in) {
    this.authTokenExpiresAt = new Date(
      Date.now() + (clientResponse.response.expires_in * 1000),
    );
    logger.info('New access token obtained :', clientResponse.response);
  } else {
    logger.error('Response lacking token TTL (expires_in)');
    throw new AuthTokenRefreshError('Response lacking token TTL (expires_in)');
  }
} else {
  logger.error('No refresh token in response :', clientResponse.response);
  throw new AuthTokenRefreshError('No refresh token in response');
}
```

By approaching it this way you can make sure the resulting thrown errors are very specific to exactly what went wrong; the larger try block means your catch block necessarily has to become more generic because it has to cover a wide variety of possible thrown errors (some of which are better passed directly with nothing in the middle)

slifty (changes requested):
Sorry for missing this originally -- let's make sure our log calls are using the right levels.

slifty (approved):
Woohoo!!!  Well done @Fenn-CS thank you for all your excellent work on this.

PR #194, opened 2023-06-29
Add custom fusion auth client

Fenn-CS:
To get a `refreshToken` with the login response we need to ask FushionAuth. The `@fusionauth/typescript-client` login method does not make that possible.

This commit adds a custom FusionAuth client that would hit the FusionAuth endpoints directly in cases where we need it.


Fenn-CS:
Not relevant anymore due to solution being found as documented here : https://github.com/PermanentOrg/sftp-service/issues/175#issuecomment-1664314321

PR #203, opened 2023-07-26
Log caught errors

slifty:
This PR updates our `catch` clauses to log any caught erros in debug.

Note that in some cases the "error" is expected to be caught and is part of intended control flows (though as I type this I realize that might be an anti pattern...).

Regardless, these logs should help us going forward.


cecilia-donnelly (approved):
Thanks!

PR #230, opened 2023-08-23
Update to node 18

### LINKED ISSUE #204: Node 16 EOL Sep 11, 2023
This project is currently using Node.js 16:

https://github.com/PermanentOrg/sftp-service/blob/4825fb281c599c59e03ac603df39fdac6d251934/.node-version#L1

Node.js is currently in maintenance LTS, and [maintenance expires September 11, 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol).

Before support runs out, we should upgrade at least to 18 (supported through April 2025), and ideally to 20 (supported through April 2026).

slifty:
This PR updates our workflows and specified node version to the current active release (Node 18)

Note that this PR should not be merged until corresponding changes to the infrastructure repository are made.

Resolves #204


.node-version
```diff
@@ -1 +1 @@
-16.17.1
+18.17.1
```
slifty, line 1: I actually think we'd be better off NOT pinning a specific minor / patch version of node here, so I'm gonna update this to just be `18`

meisekimiu (approved):

slifty:
I rebased this to main -- Since we're planning on merging the node 18 updates I'm going to merge this so it doesn't continue to fall behind.

PR #251, opened 2023-09-14
Refactor temporary storage management

### LINKED ISSUE #144: Temporary file storage can become "dead" if `CLOSE` is never called
The current implementation of upload has a DOS vulnerability: any client could open a file for write, write data (which is stored), and never resolve the file (never call `CLOSE`).

Currently this would result in a temporary file that is never deleted.

We should add some kind of timeout that will delete / close a temporary file if it hasn't been written to in a given amount of time.

slifty:
This PR refactors the way that temporary files are generated and accessed by the SFTP session handler, and incorporates functionality to "time out" temporary files that have not been accessed for a long period of time.

The PR also improves the way opened server resources are tracked.

Resolves #144

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -3,13 +3,14 @@ import path from 'path';
 import fetch from 'node-fetch';
 import { v4 as uuidv4 } from 'uuid';
 import ssh2 from 'ssh2';
-import tmp from 'tmp';
 import { logger } from '../logger';
 import { generateFileEntry } from '../utils';
+import { MissingTemporaryFileError } from '../errors';
 import { PermanentFileSystem } from './PermanentFileSystem';
+import { TemporaryFileManager } from './TemporaryFileManager';
 import type { AuthenticationSession } from './AuthenticationSession';
 import type { PermanentFileSystemManager } from './PermanentFileSystemManager';
-import type { FileResult } from 'tmp';
+import type { TemporaryFile } from './TemporaryFileManager';
 import type {
   Attributes,
   FileEntry,
@@ -22,10 +23,6 @@ import type {
 const SFTP_STATUS_CODE = ssh2.utils.sftp.STATUS_CODE;
 
 const generateHandle = (): string => uuidv4();
-
-interface TemporaryFile extends FileResult {
-  virtualPath: string;
-}
 
 enum ServerResourceType {
   Directory = 'directory',
@@ -51,7 +48,6 @@ interface PermanentFileResource extends GenericServerResource {
 
 interface TemporaryFileResource extends GenericServerResource {
   resourceType: ServerResourceType.TemporaryFile,
-  temporaryFile: TemporaryFile,
 }
 
 type ServerResource = DirectoryResource | PermanentFileResource | TemporaryFileResource;
@@ -60,6 +56,8 @@ export class SftpSessionHandler {
   private readonly sftpConnection: SFTPWrapper;
 
   private readonly activeHandles = new Map<string, ServerResource>();
+
+  private readonly temporaryFileManager = new TemporaryFileManager();
 
   private readonly permanentFileSystemManager: PermanentFileSystemManager;
 
@@ -297,7 +295,6 @@ export class SftpSessionHandler {
       'Request Data:',
       { reqId, data },
     );
-
     const serverResource = this.activeHandles.get(handle.toString());
     if (serverResource === undefined) {
       logger.verbose(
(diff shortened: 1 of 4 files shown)


src/classes/SftpSessionHandler.ts
```diff
@@ -126,9 +149,23 @@ export class SftpSessionHandler {
         length,
       },
     );
-    const file = this.openFiles.get(handle.toString());
-    if (!file) {
-      logger.info('There is no open file associated with this handle', { reqId, handle });
+    const serverResource = this.activeHandles.get(handle.toString());
```
slifty, line 149: These checks are repeated in a bunch of places.  Since most of the repetition is in the form of logs and response handling, I'm not sure if it's possible to DRY it out without also inappropriately injecting "handler" logic farther down the stack.  Still I wanted to flag it for consideration.

src/classes/SftpSessionHandler.ts
```diff
@@ -346,43 +338,74 @@ export class SftpSessionHandler {
...
+              {
+                reqId,
+                code: SFTP_STATUS_CODE.FAILURE,
+                path: temporaryFile.virtualPath,
+              },
+            );
+            this.sftpConnection.status(
+              reqId,
+              SFTP_STATUS_CODE.FAILURE,
+              'An error occurred when attempting to write to the temporary file associated with this handle.',
```
cecilia-donnelly, line 361: Thanks for these complete error messages!

src/classes/SftpSessionHandler.ts
```diff
@@ -517,63 +539,34 @@ export class SftpSessionHandler {
...
-            {
-              reqId,
-              code: SFTP_STATUS_CODE.FAILURE,
-              path: temporaryFile.virtualPath,
-            },
-          );
-          this.sftpConnection.status(
-            reqId,
-            SFTP_STATUS_CODE.FAILURE,
-            'An error occurred when attempting to register this file on Permanent.org.',
```
cecilia-donnelly: I would expect to still see this error somewhere. I'm sure I'm missing it - help?

cecilia-donnelly (commented):
I think it has to do with the change in how we represent a file, and I'm sure you explained it, but I'm confused about where files (as opposed to directories) are created on Permanent with this change. Can you point it out?

src/classes/SftpSessionHandler.ts
slifty: Oh goodness -- see this is what happens when one tries to change too many things at once and doesn't have automated tests!  Really great catch.  Fixing this now.

slifty: Status report: I've fixed this specific issue up; running some more robust manual tests though before asking for a re-review.  I'll ping later today when this is ready for you again!

src/classes/SftpSessionHandler.ts
```diff
@@ -401,74 +497,162 @@ export class SftpSessionHandler {
...
+          {
+            reqId,
+            code: SFTP_STATUS_CODE.FAILURE,
+            path: virtualFilePath,
+          },
+        );
+        this.sftpConnection.status(
+          reqId,
+          SFTP_STATUS_CODE.FAILURE,
+          'An error occurred when attempting to access the temporary file associated with this handle.',
```
slifty, line 562: (picked this instance randomly but the comment applies everywhere)

The statuses are only sent to the client, and do not appear in the logs.  This is unfortunate and will make it harder to debug (the server logs just say there is a failure, but don't indicate the reason for the failure).

slifty: (I don't think this should necessarily be addressed in this PR, since it applies to *all* failure messages in the project.)

slifty:
All righty that was gnarly.  @cecilia-donnelly I have done the following:

1. Actually implemented file upload 😐 (good gracious)
2. Rebased on main (there were a LOT of merge conflicts, since getCurrentFileSystem() had become an async method)
3. Tested uploading a file, a folder, and a file within a folder against this branch.

I think this is ready for another review.

In doing the above work I did realize two things in the main branch that I'll make issues for (a bug where the permanentfilesystem `catch` clauses could result in hung connections, and a potentially confusing implementation of `DELETE` where attempting to delete a file that doesn't exist would result in a failure)

src/classes/SftpSessionHandler.ts
```diff
@@ -404,79 +477,106 @@ export class SftpSessionHandler {
...
-              this.sftpConnection.status(
-                reqId,
-                SFTP_STATUS_CODE.FAILURE,
-                'An error occurred when attempting to register this file on Permanent.org.',
-              );
-            });
-          }).catch((fileSysErr) => {
-            logger.error(`Error loading file permanent file system ${fileSysErr}`);
-          });
+    );    const serverResource = this.activeHandles.get(handle.toString());
```
cecilia-donnelly: This might be cleaned up in the next commit but there should be a newline here.

cecilia-donnelly: (Thanks!)

cecilia-donnelly (approved):
I think this looks right. @meisekimiu is going to try it out on her local copy (thank you!).

meisekimiu (approved):

PR #296, opened 2023-10-31
Fix trivial misspelling

kfogel:


kfogel:
Hey, @Fenn-CS.  I think I made this typo a long time ago.  I just spotted it now.  I didn't want to push straight to `main` because I wasn't sure whether Permanent.org is strict about always going through a PR -- so just in case, here's a PR to review.  But it's a one-character change :-).

slifty:
Approved and merging!

kfogel:
So fast!  Thanks, @slifty.

Your reward is that I'm about to send you another PR :-).

PR #305, opened 2023-11-07
Refactor refresh tokens

### LINKED ISSUE #288: Permanent file system errors can result in hung connections
As part of https://github.com/PermanentOrg/sftp-service/pull/192 we updated the retrieval of a permanent file system to become an async method.  Part of that change involved the insertion of catches in case there is an issue accessing the file system.  Unfortunately that caught error does not result in a returned SFTP status, meaning the connection *will* hang in the event that the error is thrown.

(I'm going to blame the github UX for this -- it hid all of those changes behind a "we don't show diffs of large changes" and so unfortunately I missed that *entire file* in my review).

This is a bug! It needs to be fixed!

The fix could come in the form of adding failure status responses to those handlers, but really the mistake is that a static auth token should not be passed to the permanent file system -- the token MANAGER should be passed to the file system, and the file system should be generating the token at the moment it is needed.

This would mean the filesystem lookup becomes static again.

### LINKED ISSUE #289: Duplicated FusionAuth env vars
There seem to be two sets of env vars related to fusionauth configuration that contain the same values in practice:

1. `FUSION_AUTH_SFTP_APP_ID` and `FUSION_AUTH_SFTP_CLIENT_ID`
2. `FUSION_AUTH_KEY` and `FUSION_AUTH_SFTP_CLIENT_SECRET`

We should either (A) confirm they are potentially distinct and document as such or (B) eliminate the redundancy and just keep one of each pairing.

### LINKED ISSUE #298: 2FA auth flow does not utilize refresh tokens
It looks like #175 did not account for the [`processTwoFactorCodeResponse` flow](https://github.com/PermanentOrg/sftp-service/blob/81a84f348f737c79505986d77ef14dadb6d2ec43/src/classes/AuthenticationSession.ts#L281-L307), which means if a user has 2FA enabled then they get an auth token with no refresh token.

We'll need to refactor to have both flows set up refresh tokens properly.  This work may be done as part of #288, but I wanted to open the issue separately since they are technically separate issues.

slifty:
This PR moves refresh token logic into a dedicated class that is ONLY responsible for generating a non-stale token.

It also changes when in the SFTP flow those tokens are actually generated, resulting in a more simple SFTP flow (and also ensuring that all SFTP processes are working with fresh tokens).

One side effect of this PR is we are no longer using the `FUSION_AUTH_SFTP_APP_ID` environment variable, and are instead only using `FUSION_AUTH_SFTP_CLIENT_ID`.  These were two names for the same value, and I felt that `CLIENT_ID` was a the more consistent / clear name so decided on that one.

Resolves #298 
Resolves #289
Resolves #288


src/classes/SftpSessionHandler.ts
```diff
@@ -94,39 +94,35 @@ export class SftpSessionHandler {
...
+            'Response: Status (NO_SUCH_FILE)',
+            {
+              reqId,
+              code: SFTP_STATUS_CODE.NO_SUCH_FILE,
+            },
+          );
+          this.sftpConnection.status(reqId, SFTP_STATUS_CODE.NO_SUCH_FILE);
+          break;
+        default: {
+          this.openExistingFileHandler(
```
Fenn-CS, line 111: Question, is there a bottleneck to not awaiting this for example and making the handler method async?

Fenn-CS (approved):
Nothing looks out of place me, test works! Great work @slifty

src/classes/SftpSessionHandler.ts
slifty: We can't make the handler methods async because we don't control their signature -- the ssh2 library (which ultimately gets these handlers passed to them) expects handlers to be synchronous, which means they don't handle failed promises.

PR #307, opened 2023-11-09
Update caches on item creation

### LINKED ISSUE #306: Update cache after file / folder creation
When we were running load testing @liam-lloyd observed that the backend was getting a ton of requests to `/folder/getWithChildren` which was the primary cause of the backend buckling.

This made me realize that *because of how rclone does uploads* every time a new file is created, it will invalidates the cache for its parent folder.  Once a file is uploaded, rclone will `read` the file's stats in order to ensure it uploaded properly.  Our cache doesn't know about the file, because the file had not been read yet and therefore not loaded into the cache.

I believe that a very meaningful optimization would be to manually update the parent folder cache record to include the file, once a file is created (and likewise for a folder).

The risks here are that currently we have a simplifying assumption that the cache is returning exactly what the server response returns when loading a folder; that being said, I think it will be a safe change, and the archive record being inserted is itself the server's response to the POST.

Knock on wood, but I suspect this change will fix the scalability issues -- we would be moving from O(n calls to that very expensive endpoint to O(m) where "n" is number of files uploaded and "m" is simply number of distinct folders being uploaded *to*.

slifty:
This PR improves our caching logic so that we update our local caches to reflect the state of permanent upon archive record and archive folder creation.  In both cases this will prevent unnecessary call to the permanent back end.

I did verify locally that this lowers the number of calls to the API by 1 for each archive record being created.

Resolves #306

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -238,7 +238,7 @@ export class PermanentFileSystem {
       contentType: 'application/octet-stream',
       size,
     };
-    const archiveRecordfragment = {
+    const archiveRecordFragment = {
       displayName: archiveRecordName,
       fileSystemCompatibleName: archiveRecordName,
     };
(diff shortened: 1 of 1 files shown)


Fenn-CS (approved):
Looks good. Quite positive it would guard against other calls other than `/folder/getWithChildren`.

(Would know once I anyalize the logs from the current test runs)

PR #318, opened 2023-11-14
Bypass cache busting when opening files for write

### LINKED ISSUE #308: Consider bypassing cache buster when opening a file for write
Right now we have logic to refresh the cache if a requested path does not map to a file (or folder).

This is important for some usage patterns, but is actually counterproductive in the case that a new file is being written.  To illustrate, it would look like this:

1. Open a path for a new file (write)
2. The path doesn't exist!
3. Reload the parent folder and check again
4. The path doesn't exist!
5. Move along

As opposed to:

1. Open a path for a new file (write)
2. The path doesn't exist!
3. Move along

Again, for *read* it is useful to go through all steps because we don't want to report that a file isn't available if it actually is; but in the case of writing we're building in extra (expensive) steps.

This is somewhat related to #306

slifty:
This PR updates our cache buster so that it is generally NOT invoked when files are being opened for the purposes of writing.

It does make sure that at least the parent folder has been loaded into the cache (though this would only need to happen once per folder, which seems appropriate)

Resolves #308

### DIFF
--- a/src/classes/PermanentFileSystem.ts
+++ b/src/classes/PermanentFileSystem.ts
@@ -344,6 +344,38 @@ export class PermanentFileSystem {
     }
   }
 
+  public folderPathExistsInCache(itemPath: string): boolean {
+    if (this.folderCache.has(itemPath)) {
+      return true;
+    }
+    const parentPath = path.dirname(itemPath);
+    const folderName = path.basename(itemPath);
+    const parentFolder = this.folderCache.get(parentPath);
+    if (!parentFolder) {
+      return false;
+    }
+    const foundFolder = parentFolder.folders.find(
+      (folder) => folder.fileSystemCompatibleName === folderName,
+    );
+    return foundFolder !== undefined;
+  }
+
+  public archiveRecordPathExistsInCache(itemPath: string): boolean {
+    if (this.archiveRecordCache.has(itemPath)) {
+      return true;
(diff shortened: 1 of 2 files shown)


src/classes/PermanentFileSystem.ts
```diff
@@ -344,6 +344,26 @@ export class PermanentFileSystem {
     }
   }
 
+  public folderPathExistsInCache(itemPath: string): boolean {
+    return this.folderCache.has(itemPath);
```
slifty, line 348: This should also check the parent folder's list of folders (similar to `archiverRecrodPathExistsInCache`)

src/classes/SftpSessionHandler.ts
```diff
@@ -1108,194 +1134,180 @@ export class SftpSessionHandler {
...
-              reqId,
-              code: SFTP_STATUS_CODE.NO_SUCH_FILE,
-              path: filePath,
-            },
-          );
-          this.sftpConnection.status(
+      }
+
+      try {
+        const file = await permanentFileSystem.loadFile(filePath, true);
```
slifty, line 1265: This should possibly NOT pass `true` in the second parameter -- I think we want to load the file using cached data if cached data exists.

Fenn-CS (approved):
Nice work @slifty !

PR #329, opened 2023-11-30
Remove wait for FileVOs to be populated

meisekimiu:
Now that the back end populates FileVOs in the response of `registerRecord` (see: https://github.com/PermanentOrg/back-end/pull/525), we don't have to wait an arbitrary amount of time for processing to happen on files. Remove the code that performs this wait, since its no longer necessary and it adds a theoretical 1 second delay per file in the upload process.


slifty (approved):
I feel faster already!

PR #334, opened 2023-12-04
Pin @permanentorg/sdk to 0.6.0

liam-lloyd:
An upcoming release of @permanentorg/sdk will be breaking, but will probably not be a major version upgrade (since we're still on an unstable v0). This commit pins the version of @permanentorg/sdk used here to 0.6.0 to avoid using the updated version before this repo is ready for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
   },
   "dependencies": {
     "@fusionauth/typescript-client": "^1.48.0",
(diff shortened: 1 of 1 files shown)


liam-lloyd:
Should be merged prior to https://github.com/PermanentOrg/permanent-sdk/pull/217

PR #573, opened 2025-03-11
Update to Node 22

slifty:
This PR bumps the project to use the current active release of Node (v22).

Since Node 18 is about to leave LTS, and since the next release of the permanent SDK will require at least Node 20 this is a necessary change.


liam-lloyd (commented):
We should update the node version on the Github workflows too, right?

liam-lloyd (approved):

PR #575, opened 2025-03-12
Opt to fail on duplicate names when creating records and folders

liam-lloyd:
By default, Permanent allows the creation of folders and records with duplicate names, creating disambiguated "download names" under the hood. However, an SFTP server should not be able to create items with duplicate names, nor should it create items with different names than what it was asked to create. To this end, we added a failOnDuplicateName flag to the Permanent endpoints that create records and folders. This commit updates the SFTP service to call those endpoints with that flag set to true.


liam-lloyd:
Don't merge without https://github.com/PermanentOrg/infrastructure/pull/167

slifty (approved):
This successfully stopped SFTP from creating duplicates.

PR #715, opened 2025-09-18
Add Tests

slifty:
This is a WIP commit (all written by Claude Code but wanted to share for fun)


slifty:
I think we can all agree that *WAS* fun.

Time to delete this and actually make these tests with intention.

PR #803, opened 2026-05-04
Replace `jest` with `vitest`

slifty:
This PR replaces `jest` with `vitest` as our test runner.

Resolves #802

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
 		"format:eslint": "eslint ./src --fix || true",
 		"start": "npm run build && node build/index.js",
 		"dev": "nodemon --delay 500ms --exec ts-node src/index.ts",
-		"test": "jest",
-		"test:ci": "jest --coverage"
+		"test": "vitest run",
+		"test:ci": "vitest run --coverage"
 	},
 	"author": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -39,23 +39,22 @@
 	"private": true,
 	"devDependencies": {
 		"@tsconfig/node24": "^24.0.4",
-		"@types/jest": "^30.0.0",
 		"@types/node-fetch": "^2.6.4",
 		"@types/ssh2": "^1.15.5",
 		"@types/tmp": "^0.2.6",
 		"@types/uuid": "^11.0.0",
+		"@vitest/coverage-istanbul": "^4.1.5",
```
slifty, line 46: There are other tools for coverage calculations -- istanbul is what jest used, so keeping it here maintains line coverage parity. v8 was also on the table but I didn't feel there was much value in switching as part of this migration.

liam-lloyd (approved):

slifty:
I did look into the drop in coverage, it seems it's just that there are a few files (e.g. index.ts that just export things) that the old coverage tool counted as LoC but the new one doesn't for { reasons that seem... fine? }.

PR #806, opened 2026-05-04
Use native `fetch`

slifty:
This PR removes `node-fetch` from the repository and updates our implementation to use native node `fetch`

Resolves #805

### DIFF
--- a/src/classes/SftpSessionHandler.ts
+++ b/src/classes/SftpSessionHandler.ts
@@ -1,6 +1,5 @@
 import fs from "fs";
 import path from "path";
-import fetch from "node-fetch";
 import { v4 as uuidv4 } from "uuid";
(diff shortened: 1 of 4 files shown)


src/classes/SftpSessionHandler.ts
```diff
@@ -201,7 +200,7 @@ export class SftpSessionHandler {
 				switch (response.status) {
 					case 206:
 					case 200: {
-						const data = await response.buffer();
+						const data = Buffer.from(await response.arrayBuffer());
```
slifty, line 203: This *is* a changed line -- it should work fine, and local testing showed things still worked, but I flag it for extra attention since this line is not covered by tests.

liam-lloyd (approved):

PR #807, opened 2026-05-05
Migrate to ESM

### LINKED ISSUE #804: Migrate to ESM
A few of our dependencies are only releasing esm versions at this point, which means that we're stuck pinning older versions.  With #802 closed it shouldn't be too hard to just update to ESM.

slifty:
This PR moves the sftp service to using ESM instead of Common JS.
 
At a code level the most notable byproduct is the need to use `.js` extensions on relative imports, which is an unfortunate reality of "Typescript + ESM".
 
This change will allow us to unpin major versions of dependencies that have moved to only support ESM.

Resolves #804


package.json
```diff
@@ -48,9 +48,8 @@
 		"eslint-config-love": "^153.0.0",
 		"eslint-config-prettier": "^10.1.8",
 		"eslint-plugin-sort-exports": "^0.9.1",
-		"nodemon": "^3.1.14",
```
slifty, line 51: Love to see the removal of dev dependencies :D

liam-lloyd (approved):

